PR #27, opened 2017-06-27
fix bug of filters not being removed from the context menu

beturs:
When i removed a filter, the visual indicators, and the entry in the context menu were not removed correctly


TonyGermaneri:
bug fixed in next release

PR #135, opened 2018-06-28
Remove duplicated constants

Imperat:
There is no issue for this. I just found this place to refactoring while reading source code


TonyGermaneri:
Thanks for that @Imperat !

Issue #140, opened 2018-07-16
Add ability to insert input for edit cell in custom place of DOM


Imperat:
I use canvas data grid with fullscreen api. When my div element (parent of canvas data grid) is going into fullscreen mode the input control is going to uneditable. 
Can we add api to insert input component in custom place of dom?

fiddle to reproduce bug with fullscreen mode:
http://jsfiddle.net/sg96jL3k/

TonyGermaneri:
Thanks for the help!!  

I don't think this will break any previous implementations, but I could be wrong.

Imperat:
Hi, @TonyGermaneri. Please, have a note that fullscreen mode also affects context menu. I'm going to fix it again in weekends

TonyGermaneri:
Ah, yeah, it would.

TonyGermaneri:
Hi @Imperat I've added the event appendeditinput to accommodate your use case.  Attach to the event and run `e.preventDefault();` to stop it from adding the input to the body.  You can then add the input where you like and alter it if need be.

Imperat:
Thanks a lot, @TonyGermaneri. Now I use another event preventing for resolve this case. I forget it in today Sunday.. probably beforebeginedit :)
Tool which I implement using your library is actively used in production! Thank you again for you great library!

TonyGermaneri:
> Tool which I implement using your library is actively used in production!

That makes my day!

PR #141, opened 2018-07-16
Append input component to parent element of grid instead of body

Imperat:
Fixes issue #140 .

Instead of adding input and select element to the DOM into the end of the body I purpose to add these elements to parent element of canvas data grid. It's needed to avoid bug with fullscreen mode


Imperat:
Sorry, it's my fail :( I'll think about how to fix it

TonyGermaneri:
Damn, that's no good.  Looks like it might need a CSS reset?  I can't seem to reproduce it, do you have an example?

Issue #184, opened 2019-01-02
selectionFollowsActiveCell not reseting selectionBounds on keydown


cthurston:
### Expected behavior and actual behavior.
SelectionBounds to be in sync with selections on key navigation.

### Steps to reproduce the problem.
selectionFollowsActiveCell = true

Shift select a few cells with keyboard navigation.  Then just use the keyboard to navigate without shift or ctrl and the selectionBounds will be the previous selection.

cthurston:
Looking through the keydown event, I think the fix is as simple as changing:
````
ev = {
   selectedData: self.getSelectedData(),
   selections: self.selections,
   selectionBounds: self.getSelectionBounds()  // was self.selectionBounds
};
````

TonyGermaneri:
I see, so the event is showing a function instead of the selection bounds?  While this is getting fixed and deployed you can get at the same, and hopefully correct data using `grid.selectionBounds`.

If you are sure that fixes the problem, feel free to make a PR.  I'll fix it myself in short order if you're not feeling like it.

And thanks for finding the bug!

ndrsn:
Seems to be working.

PR #185, opened 2019-01-02
Get current selection bounds on key navigation.

cthurston:
Fixes issue #184.

- 
- 
-

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -204,7 +204,7 @@ define([], function () {
             if (!(cell && cell.header)) { return; }
             //HACK for IE10, does not like literal enum
             enu = cell.header['enum'];
-            if (self.dispatchEvent('beforebeginedit', {cell: cell, NativeEvent})) { return false; }
+            if (self.dispatchEvent('beforebeginedit', {cell: cell, NativeEvent: NativeEvent})) { return false; }
             self.scrollIntoView(x, y);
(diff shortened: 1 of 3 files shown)


cthurston:
I needed access to the NativeEvent in beforebeginedit so that I could prevent default enter from occurring in the newly created editor.

cthurston:
Noticed the controlInput was not being removed from the DOM.  Also, removed the copy/paste listeners.  I didn't test this on child grids to see if it would break anything.

lib/dom.js
```diff
@@ -204,7 +204,7 @@ define([], function () {
             if (!(cell && cell.header)) { return; }
             //HACK for IE10, does not like literal enum
             enu = cell.header['enum'];
-            if (self.dispatchEvent('beforebeginedit', {cell: cell})) { return false; }
+            if (self.dispatchEvent('beforebeginedit', {cell: cell, NativeEvent})) { return false; }
```
TonyGermaneri: Can you change this to es5 syntax?  There's no transpiler hooked up to build and IE11 is a target.  Otherwise, looks great.

cthurston: Sure thing.

TonyGermaneri:
THANKS! @cthurston Great catch and fix.  I'll get this tested and deployed today.

PR #205, opened 2019-03-27
Check for NaN on keypress navigation.

cthurston:
When a grid has extra gray space at the bottom, a user could click in that region then attempt to keyboard navigate.  This caused the rowIndex to be NaN which was throwing an error.


TonyGermaneri:
Thanks for the fix!

Issue #217, opened 2019-05-27
table data change does not change icon of column header


Hybueno09:
### Expected behavior and actual behavior.
when I click the button named _button1_ to load new data, i expecte the arrow icon is not exists.
how should I do, thanks much

### Steps to reproduce the problem.
**Example online:**  [sandbox example](https://codesandbox.io/s/cranky-ives-m4rfb)
when I click column header such as _col1_,the table column is sorted.Then I click the button named _button1_ to load new data then render it.But the arrow icon is still exists.

TonyGermaneri:
This is the same as issue #215. The problem will be fixed shortly.  Very sorry.

jswolf19:
@TonyGermaneri The same behavior occurs if a filter is applied to the data. The sort state is reset. The current sort state does not look like it is being saved in memory like the filter state is. I think the sort state should be saved, and then reapplied when necessary. #219 does not currently address the sort problem, only the filter problem, but I can look into it if you're not already working on it.

jswolf19:
@TonyGermaneri This one's going to be more difficult to do preserving current behavior while reapplying the current sort after a data reload or filter change. Currently, `order` allows a program to supply a custom `sortFunction`, so theoretically even multiple sorts on the same column can result in a different order than just applying the last sort operation. I believe this means I would need to keep all sort functions applied over the lifetime of the grid to get the same ordering, which isn't a good idea...

I want to propose keeping only the last sort per column as a first step. This will preserve the current behavior when sorting on multiple columns without worrying about keeping all of the history. I will keep track of the sort information internally for now, but it might be a good thing to expose in the public interface in the long run.

I would also propose as a further step (in the future) that the `sortFunction` argument should be dropped from `order` and only allowing it to be defined on the type or the column schema. I'm not sure what is gained by taking it as an argument, and I feel this would better unify the sorting, filtering, and formatting APIs. I will, however, honor the current APIand keep the function that was ultimately used to sort when fixing this issue.

TonyGermaneri:
I'm ok with exposing as many internals as is useful, I primarily keep stuff hidden so examination of the object isn't annoying.

I think only maintaining the last sort function for a given column is appropriate, but allowing for sorting multiple columns is important for certain data views, even though it's not supported by the UI.

What about replacing the argument with an event?  `columnsort` or something that gets called on every cell... that could be expensive.

[...]

PR #221, opened 2019-06-09
retain sort state after changing data

jswolf19:
Fixes issue #217.

- added tests around current order behavior
- fixes order when sortFunction defaults to string sorter
- fixes order to use sorter on schema if defined
- resorts after setting data
- resorts after filtering data

### DIFF
--- a/lib/intf.js
+++ b/lib/intf.js
@@ -209,6 +209,10 @@ define([], function () {
             self.resize();
             self.draw(true);
         };
+        self.applyDataTransforms = function () {
+            self.applyFilter();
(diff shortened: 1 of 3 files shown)


jswolf19:
I've run tests (passing) on Windows 10 in
- Chrome Version 74.0.3729.169 (Official Build) (64-bit)
- Firefox 67.0 (64-bit)
- Edge 42.17134.1.0
- IE 11.765.17134.0

TonyGermaneri:
Great PR!!  Thanks!

PR #236, opened 2019-08-14
Fix a bug: grid.data accidentally set to innerHTML

hackwaly:
If some style element added a text node. It will cause canvas-datagrid use interHTML to overwrite it's data.

Fixes issue #.

- 
- 
-


TonyGermaneri:
That's awesome!  I've been trying to track down why that happened.  I had like 5 commits a few months ago trying to fix it, and I ended up suppressing a parsing error message.  Thank you so much!!

Issue #62, opened 2017-11-12
PhantomJS should be replaced with headless chrome


TonyGermaneri:
I'm trying to do a perfectly legal action that works in every browser - except phantomJS.  It seems the path of least resistance here is to replace phantom.  Headless chrome (or some incarnation) seems popular.

Phantom says:
    `Attempting to change access mechanism for an unconfigurable property.`

When trying to set the property `parentNode` of the host tag <canvas-datagrid>.  `parentNode` is used as a setter internally when child grids are created.

Because of this, automated tests via command line interface no longer work.  It is necessary to run tests via an interactive browser.

PR #266, opened 2020-05-09
Replace PhantomJS with Karma (headless Chrome) for tests

ndrsn:
Fixes issue #62 

- Run tests using Karma and headless Chrome (Chromium), replacing the use of PhantomJS

I should note that the test suite produces 4 failing tests (out of 172) — unclear whether these tests were failing all along, or due to some other reason.

## Caveat

I've not been able to get a proper code coverage report out of this configuration, yet. But seeing as the last published report dates from 2017, I figured we could get away with fixing that in a second step.

Open to any suggestions / criticism you may have.


TonyGermaneri:
Awesome work!  This is badly needed to allow for the CICD pipeline to get backup and running.  I was using coveralls, but thinking using git actions might be better.  It wasn't around when I setup coveralls.

ndrsn:
Great! Yeah, I had my eye on the CI pipeline — I'll take a look at that soon.

Thanks for the quick response.

Issue #269, opened 2020-05-20
Copy/Paste on a sheet with auto generated schema gets null reference exception


iamwill:
### Expected behavior and actual behavior.
Expected:
Copy/Paste should work with an auto generated schema.

Actual:
Pasting into cell gives error in console: 
```Uncaught TypeError: Cannot read property 'name' of undefined```

### Steps to reproduce the problem.
1) Run example titled ["Create a new grid"](https://tonygermaneri.github.io/canvas-datagrid/docs/#tutorial--Create-a-new-grid)
or go to example here: http://jsfiddle.net/kvqa9ure/

2) Copy any cell

3) Try to paste contents onto another cell

PR #270, opened 2020-05-20
Fixing columnIndex key missing from schema column

iamwill:
Copy/Paste doesn't work if just providing data and no schema. If the schema is auto generated, each column is missing columnIndex which is referenced when pasting content.

Fixes issue #269

- Adding columnIndex to auto generated schema column


TonyGermaneri:
THANKS!

PR #276, opened 2020-06-19
resolve flickering cells

blacktea1105:
Flickering cells when use HTML rendering (schema data type is 'html' or set cell's innerHTML attribute in 'afterrendercell' event handler)

set grid `autoResizeColumns = true` (because value of cell width is float. Cell also flickering when set float value of column width `grid.setColumnWidth(i, 10.1)` )

see example (from 'Draw HTML via data type’ doc tutorials): https://jsfiddle.net/BlackTea1105/9kmu62jL/1/

the drawHtml function in the libs/draw.js, if cell’s size (width / height) not equals cell cache size (htmlImageCache), will clear cache and regenerate Image Element.

But the HTMLImageElement size (width and height) type always is integer, so if cell’s width is float, its never equals its cache’s width.

So I add more information into htmlImageCache (cell’s size).

### DIFF
--- a/lib/draw.js
+++ b/lib/draw.js
@@ -47,7 +47,7 @@ define([], function () {
         function drawOnAllImagesLoaded() {
             var loaded = true;
             Object.keys(self.htmlImageCache).forEach(function (html) {
-                if (!self.htmlImageCache[html].complete) {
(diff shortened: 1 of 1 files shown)


TonyGermaneri:
Very nice work!!!

PR #287, opened 2020-09-02
column width calculation error when using custom "formatter"

yuanliwei:
When using a custom formatter method, if the length of the string returned by the custom formatter method is longer than the original string length, the displayed text will be truncated.

This modification solves the problem of text being truncated by using the return value of the formatter method to calculate the width of the string.


TonyGermaneri:
Good catch, nice fix.

PR #291, opened 2020-09-04
Fill paste values in multiple selected cells

ndrsn:
Mimics Excel's paste function, where copying a single cell and selecting multiple cells and pasting results in the single value being copied over _all_ selected cells.

ezgif com-video-to-gif

This does not cover all cases handled by Excel — I will cover those in future PRs.


TonyGermaneri (approved):

TonyGermaneri:
So cool!

PR #292, opened 2020-09-05
Replace all instances of event.keyCode (deprecated) with event.key

### LINKED ISSUE #277: Use of `keyCode` is deprecated
Context: I was searching for the method that triggers when a user hits the `Delete` key on selected cells, but could not find any reference in the code base. In doing so, I discovered the `keyDown` event handler uses the key's numeric value (e.g. `9` or `27`). For readability and maintainability purposes, I had wanted to propose replacing these numeric comparisons with, for example:

```js
if (e.keyCode === keys.TAB) {
  e.preventDefault();
}
```
where `keys` is an object exported by a new `keyCodes.js` file.

_But_, in the process, I discovered that `keyCode` is considered deprecated (see [1], [2], [3])and its use is therefore discouraged in favor of `event.code` or `event.key`. I would therefore like to propose we replace the use of `e.keyCode` with `e.key` — `.key` seems to be slightly better supported by browsers (see [4] and [5])

Would a pull request implementing this proposed change be welcome?

For reference, see:

[1] https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
[2] https://developers.google.com/web/updates/2016/04/keyboardevent-keys-codes
[3] https://medium.com/@uistephen/keyboardevent-key-for-cross-browser-key-press-check-61dbad0a067a
[4] https://caniuse.com/#feat=keyboardevent-key
[5] https://caniuse.com/#feat=keyboardevent-code

ndrsn:
Fixes #277


ndrsn:
Whoops, I screwed up with an old branch... closing and reopening another

PR #297, opened 2020-09-08
Render a menu item in autocomplete filter dropdown to be able to filter blank column values

mdebrauw:
I noticed that blank cell values show up as items in the autocomplete filter dropdown menu, however they are invisible. 

Excel, for example, has a `(Blanks)` menu item which allows you filter on blank column values:

image

This PR adds a menu item (defaults to`(Blanks)` keeping in line with Excel) to the dropdown to enable filtering on blank column values (if present in the column). 

image
The value of the menu item can be customised using the attribute `blanksText`.


TonyGermaneri (approved):
Awesome work!

Issue #300, opened 2020-09-10
Copy/paste of cells in browser running on Windows does not work


mdebrauw:
### Expected behavior and actual behavior.
When copying a cell (A) (using Ctrl-c) from the canvas-datagrid and then pasting (using Ctrl-v) it into another cell (B) on the same grid I expect the value of cell (B) be the same as (A). Copying from a cell to a plain text field (or cell in edit mode) works fine.

### Specifications like the version of the project, operating system, or hardware.
Experienced this behaviour on Chrome and Firefox on Windows 10.

### Cause
On Windows the browsers deal with the clipboard data differently. So instead of getting the following from the clipboard on a paste event:

```
<meta charset='utf-8'><table><tr><td>Paste buffer value</td></tr></table>
```  

this is supplied (which looks like [CF_HTML](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/aa767917(v=vs.85)?redirectedfrom=MSDN)):

```
<html> <body> <!--StartFragment--><table><tr><td>Paste buffer value</td></tr></table><!--EndFragment--> </body> </html>
```

In the paste event handler a regex test is done on supported HTML table  structure. That test fails for the data passed from the clipboard on Windows.

PR #301, opened 2020-09-10
Fix copy/paste of cells in canvas-datagrid on Windows

mdebrauw:
This PR changes the regex checking for supported HTML to also match the format coming back from the Windows clipboard. 

It resolves issue #300.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1152,7 +1152,7 @@ define([], function () {
                 if (mimeType === "text/html") {
                     // strip table beginning and ending off, then split at rows
                     var cleanedHtmlData = data.substring(
-                        data.indexOf('<table><tr>') + 11, data.length - 13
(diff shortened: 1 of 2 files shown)


ndrsn (changes requested):
In this snippet (line 1155 of lib/events.js):

```js
var cleanedHtmlData = data.substring(
  data.indexOf('<table><tr>') + 11, data.length - 13
).split('</tr><tr>').filter(
```

the `data.length - 13` does not make as much sense any more if we're _also_ parsing CF_HTML clipboard data. To explain: that substring bit it chopping off anything _before_ and include `<table><tr>` (`+ 11`), and assuming the `data` string ends with `</tr></table>` (13 characters).

Since CF_HTML data is enclosed in `<!--StartFragment--><!--EndFragment-->`, that substring assumption no longer makes sense if you're not only parsing "regular" tables, but also parsing CF_HTML data strings.

ndrsn (approved):

PR #306, opened 2020-09-24
Fix getHeaderWidth() return value

tmaroschik:
getHeaderWidth() concatenates a string instead of summing up the column widths.

### DIFF
--- a/lib/publicMethods.js
+++ b/lib/publicMethods.js
@@ -1108,7 +1108,7 @@ define([], function () {
          */
         self.getHeaderWidth = function () {
             return self.getVisibleSchema().reduce(function (total, header) {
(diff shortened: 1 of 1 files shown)


ndrsn:
Thanks for this @tmaroschik. If you change [parseInt to include the radix](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description), I'll merge this:

> If the input string begins with "0" (a zero), radix is assumed to be 8 (octal) or 10 (decimal). Exactly which radix is chosen is implementation-dependent. ECMAScript 5 clarifies that 10 (decimal) should be used, but not all browsers support this yet. For this reason, always specify a radix when using parseInt.

PR #308, opened 2020-09-25
Fix bug that causes current selections to be overwritten when cell on only 1 row is selected

mdebrauw:
PR fixes following issue: When copy/pasting a cell from only 1 row, `affectedCells` will not get passed onto the `afterpaste` event. This is caused by the re-assignment of `self.selections` with an empty `selections` array which is only populated in the situation where we are copying multiple rows of data.


ndrsn:
Great, thanks!

PR #309, opened 2020-09-29
Fix grid not resizing when `autoResizeRows` is enabled

mdebrauw:
Trigger a resize of the grid when row height has changed due to wrapping cell contents.


ndrsn:
Ah great! We missed this earlier...

PR #310, opened 2020-09-30
Move from AMD to ES6 modules

ndrsn:
Replaces the `define(["./moduleA", "./moduleB"], function(moduleA, moduleB) { ... })` syntax with modern ES6 `import` statements. This is a first step towards eventual conversion to Typescript.

_Note:_ the diff make it appear as though a lot has changed, but that's because everything's been unindented one level. Only the imports have changed, and in `main.js` a variable was renamed to avoid collision (`component` vs `webComponent`).


jswolf19:
@ndrsn This is definitely not ES5, so from 0.25.x, I think no IE11 support, but up to v0.24.4, it seems likely to be patchable.
[see previous discussion here](https://github.com/TonyGermaneri/canvas-datagrid/commit/4221a827346f249b29c95d6bb8fa33fc25b95308#r47411914)

ndrsn:
Argh, I see, bummer — how do we patch (i.e. publish) an older release?

jswolf19:
We should be able to publish a release on a separate branch from master. We could either version as v0.24.5 or 0.24.4.1 or some other version numbering. I don't know how npm works with that, but it should be doable, as many projects support multiple versions of the same product, though I don't know if it's common with JavaScript or not...

ndrsn:
Ah, OK! That makes sense. We'll have to involve @TonyGermaneri for this, as I don't have direct access to the package repository NPM.

Are you in any position to file a PR for this?

jswolf19:
Not at the moment, but I imagine it's not too much of a priority, since no one else seems to be complaining ^_^
I should be able to get to it eventually, though we can figure out the details of how to publish separate from the PR. If you want, I can open an issue, at any rate.

jswolf19:
I've just done some very cursory testing on a branch off of [v0.24.4](https://github.com/jswolf19/canvas-datagrid/tree/v0.24.4-IE11support). I want to rearrange the commits to match the order of the fixed commits (that way it should be easier to patch other versions with rebasing, if someone wants to use an earlier version for some reason), but it seems to work in IE.

PR #317, opened 2020-11-18
Added ESModule dist

### LINKED ISSUE #316: Isolated Instances
Is there any way to isolate the instance of canvas-datagrid?
In large apps, each section needs to be able to load and unload its dependencies since other portions of the app often have no need for most of them.
Having some way to keep it out of global scope would be helpful.

josh-hemphill:
Closes #316 

- Added ES Module in dist
- Only added rollup for ES Module support (webpack still does not; webpack config could be converted to rollup)
- ES Module does not include global variables, but all others are unchanged.

It would be trivial to migrate to Rollup instead of upgrading to Webpack 5, if there's interest in doing so.
I also noticed there were some out of date dependencies, should I create a separate PR for that?


josh-hemphill:
@TonyGermaneri Would you rather me look at migrating the webpack to rollup to keep from splitting the config?

ndrsn:
Hi @josh-hemphill, fellow maintainer here — thanks for this pull request! It seems that with the new release of Webpack (version 5), ESM _are_ supported (from https://github.com/purtuga/esm-webpack-plugin):

>  Webpack 5 has been released and with it comes support for output ES Modules. See documentation here. 

It's not obvious to me from a quick glance at the docs, but [it looks like this PR](https://github.com/webpack/webpack/pull/10953) addresses the topic? Would you be interested in modifying this PR to upgrade to Webpack v5 for generating ESM, instead of using Rollup?

(I should add I have nothing against Rollup _per se_), but unless there's a strong argument for switching, I believe it'd be better to stick with Webpack, considering that's what we're both already more familiar with.

josh-hemphill:
@ndrsn I first tried getting webpack to do it natively, since there's the experimental features for `.mjs` support, but apparently it's [not in a working state](https://github.com/webpack/webpack/issues/2933). (I spent a lot of time on that, and couldn't get it working)
[The plugin you linked](https://github.com/purtuga/esm-webpack-plugin) is actually locked to Webpack v4 because they thought v5 was supposed to have module support. [See their issue](https://github.com/purtuga/esm-webpack-plugin/issues/21#issuecomment-715753388)
The [exports PR](https://github.com/webpack/webpack/pull/10953) is about supporting ESM as a dependency of your package, reading that information from `package.json`s of your dependencies.
For the ESM support, Rollup has the advantage that it's tree-shaking after making all the global assignments unreachable, which removes all that code from the bundle.  ESM support is very stable in rollup, and it's pretty fast (not quite as fast as snowpack, but faster than webpack). I haven't actually tested it, but recreating the webpack config in rollup should look and operate almost the same.

ndrsn:
@josh-hemphill ahh, thanks for that, I didn't read closely enough! So I see, it looks like they're working on [adding support to webpack soon](https://github.com/webpack/webpack/issues/2933#issuecomment-715903657), but it's not supported yet.

Alright, so we can either a) keep the current webpack config and add your Rollup config to support ES modules for the time being, or b) move everything over to Rollup.

Out of prudence I'd be inclined to go with option `a)` for now (this PR as it is now) and configure webpack for ESM once support lands. It's low-risk, and doesn't require any work from our end. But I'd be interested to hear what @TonyGermaneri  thinks?

As for the rest of the PR: your editor seems to use 4-space indents, causing all of `package.json` to be reformatted — can you reformat to 2 spaces (also for `rollup.config.js`, so we can get a more useful diff?

ndrsn:
Also:
> I also noticed there were some out of date dependencies, should I create a separate PR for that?

Which dependencies would you suggest we upgrade, and what issues would upgrading solve, if any? Obviously, any improvements are welcome :) But either way, yes, a separate PR would be more appropriate for that.

TonyGermaneri:
I would really like to keep progress going so using both for now is fine.  Also, less work is always better than more. 

A lot of the time trying to get ahead of things just ends up being premature optimization that is in reality unnecessary risk.

package.json
```diff
@@ -77,11 +82,31 @@
...
     "prettier": "^2.1.2",
+    "rollup": "^2.33.3",
     "should": "^11.2.1",
     "webpack": "^4.44.1",
     "webpack-cli": "^3.3.12"
   },
   "dependencies": {
     "is-printable-key-event": "1.0.0"
+  },
+  "exports": {
```
ndrsn, line 93: I'm having a little trouble wrapping my head around all these near-identical lines and squaring it with the documentation on [conditional exports](https://nodejs.org/api/packages.html#packages_conditional_exports). My (uninformed) spider-sense tells me we can do with less, but please correct me if I'm wrong!

Just to make sure I actually understand this: the `development` and `production` keys are to define separate package exports depending on whether `require('canvas-datagrid/development')` is called, or `require('canvas-datagrid/production')` (or the `import` equivalent) ? Do we need to explicitly define a `production` key — as in, is that not the default, on lines 106-110?

ndrsn:
> I would really like to keep progress going so using both for now is fine. Also, less work is always better than more.
> A lot of the time trying to get ahead of things just ends up being premature optimization that is in reality unnecessary risk.

Yes, agreed! So I'd like going with this Rollup config and leaving the existing Webpack as-is. @josh-hemphill , if you could address the conditional exports comment, then I think we'd be ready to merge this.

package.json
josh-hemphill: When writing that I was referencing the Webpack documentation on the subject. I don't think they create separate exports since they don't include path characters, downstream Webpack probably uses it for extended conditionals, but yes I think they're probably redundant. Except I think I'd move the development down so if someone's using webpack with script or require and are running in development mode, the `.debug.js` will be used.

ndrsn:
Merged! 🎉  Thanks for this @josh-hemphill. I'll put out a release soon.

PR #318, opened 2020-11-21
ESModule fix

josh-hemphill:
Fixes ESModule Error in #317

- Fixes code substitution so that only the global declaration is removed


josh-hemphill:
@ndrsn The ESModule doesn't work without this fix, I'd like to get it added.
I didn't test the ES Module bundle correctly in #317, would you happen to have any ideas about how to include tests for ES Modules? I couldn't find a way to add it without breaking the tests.

ndrsn:
@josh-hemphill apologies, work got a bit busy — I'll take a look at this tomorrow.

I do believe, however, that there must a better way to leave out a conditional than replacing the #NO_GLOBAL? Some kind of variable injection? I'll look into this tomorrow. Thanks for the heads-up!

ndrsn:
Can we use [Rollup's environment values flag](https://rollupjs.org/guide/en/#--environment-values) instead of the replace? For example (untested):

```
$ rollup -c --environment
```

and

```
const excludeGlobal = process && process.env && process.env.EXCLUDE_GLOBAL;

if (window && !window.canvasDatagrid && !window.require && !excludeGlobal) {
  window.canvasDatagrid = function (args) {
    return new Grid(args);
  };
}
```

rollup.config.js
```diff
@@ -28,16 +28,7 @@ export default {
...
-        {
-          match: /lib(\/|\\)main/,
-          test: /(\/\/ #NO_GLOBAL)(.|\r|\n)*\1/,
-          /* export default function canvasDatagrid (args) {
-				  return new Grid(args);
-				}; */
-          replace: ``,
-        },
-      ],
+      'window.EXCLUDE_GLOBAL': true,
```
ndrsn, line 31: I think `true` needs to be quoted, no? My linter complains:

rollup_config_js_—_canvas-josh

If you fix this, and add a comment to `lib/main.js` explaining what the use of `EXCLUDE_GLOBAL` is for (i.e. exporting as an ES module, instead of setting on `window`), I'll merge this.

ndrsn:
> I didn't test the ES Module bundle correctly in #317, would you happen to have any ideas about how to include tests for ES Modules? I couldn't find a way to add it without breaking the tests.

This is a good point: right now the tests are browser based (Karma), and [support for ES modules in browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) isn't complete I believe.

Right now I'd say don't worry about it — if the current Rollup config works, it works. But what we could do is add another, NodeJS-based test runner, to test _just_ the ES module. We need to clean/split up some of the tests any way, so I'll see if I can add this in with that.

rollup.config.js
josh-hemphill: The documentation isn't clear, it says any primitives should be valid, but it looks like they really want everything to be `JSON.stringify()`ed ideally. I'll add it that way.
Just confirmed that it does get tree-shaken with it being stringified.

josh-hemphill:
I did some manual tests in the browser with the ES Module to be safe, it all seems to be good.

rollup.config.js
ndrsn: Ah, I guess that's maybe the type definition, then. This is better, but could you change `JSON.stringify(true)` to `'true'` (i.e., just quoting it as a string)?

The former seems needlessly complex in the event of a simple plain value like `true` :-)

ndrsn:
Thanks, @josh-hemphill — I've merged this and the other changes and released it as v0.25.2.

Issue #311, opened 2020-10-03
CSP violation: Unsafe-eval


Dassderdie:
### Expected behavior and actual behavior.
I'd like to use this library without having to weaken the Content Security Policy of our application.

It seems like `eval` is used in the source code of this library (`component.js` line 150, `main.js` line 25). This isn't a good practice because of the potential XSS-vulnerabilities that come along with it.

On this note I would like to see information about the security against XSS-vulnerabilities of this package in the `README.md`.

### Steps to reproduce the problem.

Our CSP: 
```
default-src 'self';script-src 'self';frame-src 'self' blob:;frame-ancestors 'self';worker-src 'self';object-src 'none';font-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: blob:;media-src 'self' blob:;upgrade-insecure-requests
```

The Error:
```
Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".
...
    at window.canvasDatagrid.window.require.window.canvasDatagrid (canvas-datagrid.js:1)

```

Dassderdie:
@ndrsn Any update on this?

TonyGermaneri:
eval shows up in two places in the source [1](https://github.com/TonyGermaneri/canvas-datagrid/blob/afb1ab15577457599b8b6bbc76c7d35e7a5195e9/lib/component.js#L175)
[2](https://github.com/TonyGermaneri/canvas-datagrid/blob/afb1ab15577457599b8b6bbc76c7d35e7a5195e9/lib/main.js#L39)

Both of these can be replaced with a function constructor.  The constructed function can then be invoked immediately after definition.  It's a small transparent change.

In the first case, it prevents ES6 class constructor syntax, which is required to instantiate a web component, from causing syntax errors in ES5 browsers.  We can 1) drop support for es5 browsers (ideal), 2) create two different build artifacts (hard) or 3) allow function constructor to hide the ES6 syntax still.

In the second case, it is creating a proxy for the internal event hook for the web component.  This can likely be replaced with another pattern if the function constructor is undesirable from a security point of view.

In all cases we can remove eval().  From a security point of view, there is no risk here, but I understand the need to streamline the DAST process.

PRs are always welcome!

Dassderdie:
[...]

PR #319, opened 2020-11-21
Example Rollup Migration and Dependency Update

josh-hemphill:
In order to get Rollup working with everything I ended up needing to upgrade the dependencies, which led me down a rabbit hole. Feel free to merge the upgrades and QOL stuff into a different branch

- Removed unused dependencies: Chai, karma-chai, should, (webpack/webpack-cli for migration)
- Updated all dependencies
- Upgraded testing libraries (stopped security warnings)
- Moved movable config to it's own directory
- Dropped `canvas-datagrid.debug.js` as it's only difference was having a source-map, which doesn't effect performance to just include for the production bundle.
- Removed unused test-runner files, and fixed karma tests.html to remove unused and breaking imports.

Funny enough, all the time it took to do this was in upgrading the testing libraries, there were a bunch of duplicate imports etc. that were nearly impossible to diagnose. 

Contributes to #311 (Changing Eval to `Function()` was trivial, so added it)
I looked at going ahead and replacing the other Eval while I was at it, but I couldn't figure it out at the time.


ndrsn:
Hi @josh-hemphill, thanks for this. However, I feel there's too much going on for it rationally be a single pull request — if you don't mind, I'd like to suggest splitting this PR up into more distinct chunks (in whichever order makes the most sense):

- Removed unused dependencies
- Remove unused testing files
- Upgrading dependencies
- Upgrading testing libraries
- Fixing #311 

Smaller is better, because it makes the risks more manageable, and @TonyGermaneri  and mine work easier (and thus faster!).

I am not in favor of moving configuration files a separate `config/` directory: I'd prefer to follow convention instead of having to explicitly pass in a config location. Also, it makes editor tooling (prettier, etc) work out-of-the-box.

Also, we'd like to stick with our existing webpack configuration for the foreseeable future, unless a strong argument can be made for migrating to another bundler.

josh-hemphill:
I'll work on break it apart into separate PRs, though the dependency and testing upgrades might be interdependent. 

On bundlers, Rollup is usually recommended for libraries since their niche is getting really small single bundles, whereas Webpack's strength is it's codesplitting, which is ideal for websites. The plugins used (e.g. Babel and Terser) are really agnostic of which one is used, and in this case, both Webpack and Rollup are just runtimes to push everything through Babel and Terser. Though with Rollup, it might be interesting to look at creating a bundle with the extra Babel output step that can insert the Babel runtime pollyfills for anyone that need compatibility with much older browser at the cost of a larger file size.

PR #322, opened 2020-11-25
No eval

### LINKED ISSUE #311: CSP violation: Unsafe-eval

josh-hemphill:
Fixes #311 

- [x] Replace function created from `eval` with `Function()` constructor
- [ ] Replace class obfuscation created with `eval`?

I haven't been able to decipher what the other eval is actually doing to be able to look at replacing it.


package.json
```diff
@@ -1,109 +1,109 @@
 {
-  "name": "canvas-datagrid",
```
ndrsn, line 2: Indentation here is off, again — can you restore the original indentation, so the diff shows properly?

webpack.config.js
```diff
@@ -1,4 +1,5 @@
 const path = require('path');
+const webpack = require('webpack');
```
ndrsn, line 2: Why do we need to import webpack? I'm not sure it's needed for the config file — it's not being used anywhere?

esFill.js
```diff
@@ -0,0 +1,8 @@
+const exports = {
```
ndrsn, line 1: What does this file do?

ndrsn (changes requested):

webpack.config.js
josh-hemphill: Sorry it's a fragment, I had to re-delete the webpack ESM export that I had experimented with.

esFill.js
josh-hemphill: Yep, sorry, another fragment, I made a branch off of an older commit.

ndrsn:
Thanks for this, @josh-hemphill !

PR #323, opened 2020-11-25
Unused Test Files

josh-hemphill:

- Removes `test/index.js` and `test/mocha.opts`, doesn't get loaded by karma


esFill.js
```diff
@@ -0,0 +1,8 @@
+const exports = {
```
ndrsn, line 1: Same as for other PR: what does this file do, and why is it included in the pull request?

package.json
```diff
@@ -1,109 +1,109 @@
 {
-  "name": "canvas-datagrid",
```
ndrsn, line 2: Same: please fix your indentation/formatting settings to match the file's.

webpack.config.js
```diff
@@ -1,4 +1,5 @@
 const path = require('path');
+const webpack = require('webpack');
```
ndrsn, line 2: Same question: why do we need to require webpack?

ndrsn (changes requested):

PR #324, opened 2020-11-26
Unused deps

josh-hemphill:

- Remove `chai` (tests are using their own assertion functions)
- Remove `karma-chai`
- Remove `Should` (Neither `Should` nor the newer `Chai/Should` are used)


ndrsn:
Thanks for this, this feels good to clean up. As for the custom assertion functions, I do believe we should eventually use a more standard assertion library like `chai` or `assert` (my personal preference) — but that's something we can tackle at a later date.

PR #331, opened 2020-12-07
fix: use 'selectNone' on escape so selectionchanged event will be dis…

twojtylak:
When clearing the selection via Escape there won't be a `selectionchanged ` event dispatched. 

By calling  `selectNone `, analogous to `ctrl + a` with  `selectAll `, the missing event will be dispatched.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1239,8 +1239,7 @@ export default function (self) {
     }
 
     if (e.key === 'Escape') {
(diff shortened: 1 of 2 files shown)


ndrsn:
This is excellent, @twojtylak. Thank you!

ndrsn:
Released this in v0.25.4

PR #333, opened 2020-12-11
Types Definition

### LINKED ISSUE #155: Typescript typings?
It would be nice to have Typescript typings for your component.  This helps with intellisense and also with the stricter type checking.  

I've created a file and could to a PR.  It would be just a matter of distributing the file along with your component in the root folder (where package.json exists).

I created the file partly by looking at the defaults for the params but mostly by writing a quick bit of javascript in the dev tools to pull stuff out of the API documentation page.  So, not everything is guaranteed to be correct, but it would be a good starting point.

There's also JSX setup, so you don't have to use React.createElement when using React.

josh-hemphill:
Fixes #155

Finally figured it out. The answer was [tsd-jsdoc](https://github.com/englercj/tsd-jsdoc#readme)! It uses the type information returned by jsdoc to create a `.d.ts` file.

If anyone knows of a TS project using this library, I think the first step to using this would be to have them manually reference the definitions file to test it. That way we can publish without linking it in package.json and avoid breaking anything but still let people use it.

- Add type definitions for `canvas-datagrid.js`


ndrsn:
Great! Could you fix the indentation on `package.json` so the diff is more useful?

PR #334, opened 2020-12-14
Split original source data from viewed data

ndrsn:
Provides a (possible) solution for #241, by making a distinction between the data that is currently being rendered (due to filtering/sorting), and the entire set of data available to the grid.

- Adds a (internally and externally accessible) property `viewData`, which refers to the _currently rendered data_.
- Adds `boundData` which refers to `self.originalData`. Not sure if we should keep this, or just provide access to `originalData` directly.
- Changes to `cell` object (as created in `drawCell` and emitted in events), see below.

#### Changes to `cell`

The `drawCell` function constructs a cell object with a number of properties, such as style, value, but also location data. This PR:

- Keeps `rowIndex` and `columnIndex` unchanged for now, although I think we'd want to deprecate those properties to avoid confusion.
- Adds `viewRowIndex` and `viewColumnIndex` to a cell's draw to refer to the position of the cell _as it is rendered_, i.e. after filtering and sorting the data.
- Adds `boundRowIndex` and `boundColumnIndex`, which is the position in `self.originalData`, i.e. _before_ the data is sorted or filtered.

#### How it works, sort of

The new method `refresh` makes a shallow copy of `originalData`, and maps each row to a `(row, rowIndex)` tuple. It then applies filters and sorts the data in this array. After unpacking the tuples (remapping it to just rows), the resulting array is then basically `viewData`; the original row indexes (which are no longer in order due to filtering/sorting) are then added to a `Map`, which maps `viewRowIndex` to `boundRowIndex`, so we can determine the original row. This is used in `getBoundRowIndexFromViewRowIndex`

### DIFF
--- a/lib/intf.js
+++ b/lib/intf.js
@@ -1337,7 +1337,7 @@ export default function (self, ctor) {
       // apply filter, sort, etc to incoming dataset, set viewData:
       self.refresh();
       // empty data was set
-      if (!self.schema && (self.data || []).length === 0) {
+      if (!self.schema && (self.originalData || []).length === 0) {
         self.tempSchema = [{ name: '' }];
       }
       self.fitColumnToValues('cornerCell', true);
--- a/lib/intf.js
+++ b/lib/intf.js
@@ -37,6 +37,10 @@ export default function (self, ctor) {
       });
     },
     sort: function () {
+      console.warn(
+        'grid.orderings.sort has been deprecated. Please use grid.refresh().',
+      );
+
       self.orderings.columns.forEach(function (col) {
         self.viewData.sort(col.sortFunction(col.orderBy, col.orderDirection));
       });
@@ -184,11 +188,11 @@ export default function (self, ctor) {
     return selectedData;
   };
   self.getBoundRowIndexFromViewRowIndex = function (viewRowIndex) {
-    if (self.boundRowIndexMap) {
-      return self.boundRowIndexMap[viewRowIndex];
+    if (self.boundRowIndexMap && self.boundRowIndexMap.has(viewRowIndex)) {
+      return self.boundRowIndexMap.get(viewRowIndex);
     }
 
-    return -1;
+    return undefined;
   };
   self.getColumnHeaderCellHeight = function () {
     if (!self.attributes.showColumnHeaders) {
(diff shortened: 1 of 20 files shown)


lib/intf.js
```diff
@@ -267,26 +290,65 @@ export default function (self, ctor) {
...
-    Object.keys(self.columnFilters).forEach(function (filter) {
-      var header = self.getHeaderByName(filter);
+  self.getFilteredAndSortedViewData = function (originalData) {
+    // We make a copy of originalData here in order be able to
+    // filter and sort rows without modifying the original array.
+    // Each row is turned into a (row, rowIndex) tuple
+    // so that when we apply filters, we can refer back to the
+    // row's original row number in originalData. This becomes
+    // useful when emitting cell events.
+    let newViewData = originalData.map((row, originalRowIndex) => [
```
TonyGermaneri, line 300: What happens when data contains 10MM rows?  Are we making a copy of 100% of that data?  Is there a way to do this without copying the data?  Making a difference somehow?  I think this will cut performance or at least memory usage in half for larger record sets.  Shouldn't be hard to see what sort of overhead we're adding here.

lib/intf.js
```diff
@@ -267,26 +290,65 @@ export default function (self, ctor) {
...
+        ]),
+      ),
+    };
+  };
+  self.refresh = function () {
+    const { viewData, boundRowIndexMap } = self.getFilteredAndSortedViewData(
+      self.originalData,
+    );
+
+    self.viewData = viewData;
```
TonyGermaneri, line 346: `viewData` or.... `view`?  Since `view` isn't taken and `viewData` is a mouthful what do you think about changing this to `view`?

lib/intf.js
ndrsn: It shouldn't be copying the data: the `map` creates a new array, but `row` in `[row, originalRowIndex]` refers to the same object (i.e. deep-equals) as in the `originalData` array.

There _is_ a slight increase in memory usage, as we're keeping track of `originalRowIndex`, and this is then stored in the `boundRowIndexMap`, but that's `O(2N)` (~8 bytes for the `viewRowIndex` as key, and ~8 bytes for the value), so I think that's acceptable? If I find time later, I'll see if I can set up a benchmark.

lib/intf.js
ndrsn: I tend to err on the side overly verbose variable names, in case I forget three weeks down the line what it was I was doing. To me `view` is too vague in that it doesn't quite describe the nature of the variable. I agree that `viewData` is a mouthful, but feel it's easier to understand `grid.viewData[0]` as referring to the first row than `grid.view[0]` is.

PR #369, opened 2021-08-17
Fix autoscroll issue when selecting a cell on top or bottom row

mdebrauw:
Fix issue where selection of 1 or more cells on top or bottom row of the grid (once the grid has been scrolled down) adds addtional cells to the selection.

selection-top-row-after-scroll

Steps to reproduce:
- scroll down in the grid (on a scrollable canvas)
- click a cell on the top row, click in the upper half of the cell
- notice the grid scrolling down 1 or 2 rows and increasing the selected area.

I've tracked the issue down to a call of [autoScrollZone](https://github.com/TonyGermaneri/canvas-datagrid/blob/7bb0e553075f5cddfb5e70f7b6f2b9b1de813a32/lib/events.js#L557) in the [mousemove](https://github.com/TonyGermaneri/canvas-datagrid/blob/7bb0e553075f5cddfb5e70f7b6f2b9b1de813a32/lib/events.js#L408) event handler. This method [updates](https://github.com/TonyGermaneri/canvas-datagrid/blob/7bb0e553075f5cddfb5e70f7b6f2b9b1de813a32/lib/intf.js#L414) [scrollBox.scrollTop](https://github.com/TonyGermaneri/canvas-datagrid/blob/7bb0e553075f5cddfb5e70f7b6f2b9b1de813a32/lib/intf.js#L1555) and causes the grid to auto scroll down a row and retrigger `mousemove` to select additional cells. 

I've made the call conditional on `self.rowBoundaryCrossed` being `false`. This seems to make sense as we don't want to scroll the grid if the mouse moves laterally along the same row (i.e. mouse button release was delayed or user intended to select multiple adjacent cells from the same row).

selection-top-row-after-scroll-after-change

Other approaches:
- Use the `delta` variable to check that mouse didn't move; this is too brittle IMHO as the mouse can move very subtly between mouse button press and release)
- Use `self.cellBoundaryCrossed` to check that mouse cursor is till in same cell; this does not allow selection of multiple cells on the top or bottom row.

A workaround that seems to work for us is to set the the attribute `selectionScrollZone` to 0, this avoids the scroll triggered by `autoScrollZone`.

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -19,6 +19,8 @@ export default function (self) {
       ['autoGenerateSchema', false],
       ['autoResizeColumns', false],
       ['autoResizeRows', false],
+      ['autoScrollOnMousemove', false],
+      ['autoScrollMargin', 5],
       ['blanksText', '(Blanks)'],
       ['borderDragBehavior', 'none'],
       ['borderResizeZone', 10],
--- a/lib/docs.js
+++ b/lib/docs.js
@@ -25,6 +25,8 @@
  * @param {boolean} [args.editable=true] - When true, cells can be edited. When false, grid is read only to the user.
  * @param {string} [args.borderDragBehavior='none'] - Can be set to 'resize', 'move', or 'none'.  If set to 'move', `allowMovingSelection` should also be set true.  If set to 'resize', `allowRowResizeFromCell` and/or `allowColumnResizeFromCell` should be set true.
  * @param {boolean} [args.autoGenerateSchema=false] - When true, every time data is set the schema will be automatically generated from the data overwriting any existing schema.
+ * @param {boolean} [args.autoScrollOnMousemove=false] - Only trigger auto scrolling when mouse moves in horizontal or vertical direction.
+ * @param {number} [args.autoScrollMargin=5] - Number of pixels of mouse movement to trigger auto scroll.
  * @param {boolean} [args.allowFreezingRows=false] - When true, the UI provides a drag-able cutter to freeze rows.
  * @param {boolean} [args.allowFreezingColumns=false] - When true, the UI provides a drag-able cutter to freeze columns.
  * @param {boolean} [args.allowColumnReordering=true] - When true columns can be reordered.
@@ -54,6 +56,7 @@
  * @param {boolean} [args.showRowNumberGaps=true] - When true, gaps between row numbers due to filtering are shown in the row headers using a colored bar.
  * @param {boolean} [args.showRowHeaders=true] - When true, row headers are shown.
  * @param {number} [args.reorderDeadZone=3] - Number of pixels needed to move before column reordering occurs.
+ * @param {number} [args.selectionScrollZone=20] - Number of pixels to auto scroll in in horizontal or vertical direction.
  * @param {boolean} [args.showClearSettingsOption=true] - Show an option on the context menu to clear saved settings.
  * @param {string} [args.clearSettingsOptionText='Clear saved settings'] - Text that appears on the clear settings option.
  * @param {string} [args.showOrderByOptionTextDesc='Order by %s ascending'] - Text that appears on the order by descending option.
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -665,7 +665,7 @@ export default function (self) {
     self.eventParent.addEventListener('mousedown', self.mousedown, false);
     self.eventParent.addEventListener('dblclick', self.dblclick, false);
     self.eventParent.addEventListener('click', self.click, false);
-    self.eventParent.addEventListener('mousemove', self.mousemove);
+    document.body.addEventListener('mousemove', self.mousemove);
     self[self.isChildGrid ? 'parentGrid' : 'eventParent'].addEventListener(
       'wheel',
       self.scrollWheel,
--- a/lib/events.js
+++ b/lib/events.js
@@ -554,7 +554,7 @@ export default function (self) {
             self.selectArea(sBounds, true);
           }
         }
-        self.autoScrollZone(e, x, y, ctrl);
+        if (self.rowBoundaryCrossed) self.autoScrollZone(e, x, y, ctrl);
       }
     }
     self.cellBoundaryCrossed = false;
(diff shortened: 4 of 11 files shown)


TonyGermaneri:
If I read this correctly it looks like you're disabling autoscroll during selection.  I think that really needs to happen, but I agree the current behavior is kinda jankey.  I think the proper thing to do here is make the scroll engage when it hits some value above the cell instead of within the cell.  Disabling it completely removes a feature that some people need.  If you can make it more like google sheets or excel that would be much better.  I think the change would be pretty simple.

mdebrauw:
Hi, yes, I see. Agree this is not a good solution. I've pushed an update:

Google Sheets (and Excel) scroll up/down (to the invisible cells) once you've moved the mouse across the grid boundary. 

google-sheets-selection-example

As far as I can tell, within CDG we cannot detect that the mouse has moved outside the grid. Once the mouse moves out of the grid the `mousemove` event is no longer fired. 

_Am I missing something here?_

If we indeed can't detect mouse movement outside of CDG, a solution could then be to:
- On starting the selection detect the direction of mouse movement
- If the mouse moved by more than a specified margin this would trigger the autoscroll

This solution is little fiddly as scrolling stops once the mouse is moved out of the grid, it relies on keeping the mouse towards the boundary of the cell/grid

cdg-selection-example

If you think this is the right direction I will make two additional changes:
- Make this behaviour configurable through a parameter
- Make margin by which mouse has to move configurable through a parameter

TonyGermaneri:
Great research!  Checked the code, it looks like the mousemove event is indeed bound to the canvas element.  Should it be?  Any reason we can't bind it to the document?  Seems like we might get the desired behavior that way.  Shouldn't be too much different from binding to the canvas element I think.  I might be terribly wrong.

Bindings happen about here
https://github.com/TonyGermaneri/canvas-datagrid/blob/master/lib/dom.js#L668

Mind that the canvas renders child canvases recursively.  I don't predict an issue here but it looks suspect.
https://github.com/TonyGermaneri/canvas-datagrid/blob/master/lib/dom.js#L644

This is really great work you're doing!

mdebrauw:
Hi, thanks for the encouragement :-) 

I have tried your suggestion and it seems to work:

scroll-outside-grid

TonyGermaneri:
That behavior is so much better.  I was looking at the code, it looks like some of the focus setting code was removed, this might affect child grids.  Have you tested the behavior in a child grid?  I think if it works with the two types of child grid this is awesome.  There are tests that cover child grid, but I don't think they specifically cover auto scrolling on the child grid.  Let me know if you need direction on how to test this.

mdebrauw:
Ok, not too familiair with child grids, but have done some visual testing and it seems to work. Note that I've included a option to maintain the original behviour (see below).

I guess we could make this into a couple of test cases, but still wrapping my head around the way the tests have been setup. Some direction there would be helpful! :-)

### Up/down scrolling parent and child grid:

up-down-parent

up-down-child

### Sideways scrolling parent and child grid:
sideways-parent

sideways-child

TonyGermaneri:
Really great work!

jswolf19:
Sorry to dig this up after closing, I was looking at upgrading and the binding of mousemove to document.body here bit me pretty hard. It gets really crazy when you have multiple grids on the same page. Also, I believe as it's currently implemented, if the lifetime of the grid is shorter than the lifetime of the document, then this is the source of a memory leak, as well...

In general, I'm not sure how desirable it is when the user of the grid adds a mousemove event to the grid that it gets triggered for all movement even when the grid is not the target.

Is it possible to separate this internal logic that requires events from outside the grid from the mousmove exposed to users of the grid?

mdebrauw:
Hi @jswolf19,

Thanks for your feedback! Sorry to hear you had issues with this change. 

Would you be able to provide us with a Codepen or JSFiddle example of how you are using canvas datagrid.

Alternatively, if you have thoughts on how we can resolve this / support both behaviours, maybe you can provide a pull request or we can hash something out together.

Cheers, Mark

jswolf19:
@mdebrauw I basically have a datagrid, and then for certain cells with detailed info, I display that in a popup datagrid when the cell is hovered over. I use mousemove to determine the cell hovered over, the mouse position, and to let me know that I should hide the popup.

I've tried to get the important bits, along with some diagnostic addons. This is with [version 3.7](https://jsfiddle.net/jswolf19/yxoLsc1k/), and this with [version 3.8](https://jsfiddle.net/jswolf19/vhxsfLy5/).

My issues are
1. with ~v3.7, I only get a `mousemove` when the top-most element under the mouse pointer is the grid that the handler was added to, but with v3.8+, I get a `mousemove` when the mouse moves anywhere in the document. I would argue this goes against the usual `mousemove` event, which is usually only triggered "when a pointing device (usually a mouse) is moved while the cursor's hotspot is inside it." ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousemove_event))
2. with ~v3.7, the detail grids are only displayed until the mouse moves over the grid again, and then are detached from the DOM (or I'm intending to detach them, at any rate), but the `mousemove` event no longer fires on the removed grids. With v3.8+, the mouse `movemove` handlers are still bound to the body `mousemove` event even after the element is no longer visible.

As for how to resolve the issue, I think it depends heavily on how easy it is to isolate the logic for handling autoscrolling, but in my mind, the body `mousemove` event should only be bound to while the mouse is being dragged, and the handler should be removed once the drag is over (add `mouseup` on body as well and then remove both or check the `MouseEvent.button(s)`?) 

This would make it harder to use `preventDefault` to cancel the autoscroll behavior, although I'm not sure how easy that is to do currently, either. It may be worth adding a `startdrag`/`enddrag` or something similar... That may even be useful for something else I would find useful, handling dragging data from one grid to another.

faimaklg:
Oh! i think it`s bad modification. Always trigger drawing when the mouse is not in the grid, this is very inappropriate.

mdebrauw:
Hi @jswolf19 I will have a look at this. Best approach for now might be to revert the change on event bindings and come up with a different solution to support the desired scrolling behaviour.

jswolf19:
@mdebrauw I'm still using 0.22.12 at the moment, so I'm personally not looking for an immediate fix. If there's not a lot of push back (and there doesn't seem to be), then I'm not sure how necessary a reversion is, but I do think that an alternative solution should be pursued. I don't know how actively I can contribute, but I'm willing, if not necessarily able ^_~

PR #370, opened 2021-08-19
Fix paste in filter mode

mdebrauw:
When pasting data into a grid that has active filters: 

- Updates of cells don't happen correctly
- Updated cells disappear from filter results if values no longer pass filter criteria; this leads to a confusing user experience as changed cells "all of a sudden" disappear (e.g. Excel and Google Sheets maintain changed values in the filter results).

paste-when-filtered-before

- The `rowIndex` of changed cells is mapped to the `boundRowIndex` so changes apply to `originalData` correctly.
- Don't refresh the grid after a paste event.

The result after changes applied:  

paste-when-filtered-after

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1544,7 +1544,12 @@ export default function (self) {
           newRowData[columnName] = cellData;
         }
 
-        self.originalData[realRowIndex] = newRowData;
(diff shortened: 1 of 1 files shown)


TonyGermaneri:
Great fix!  I had to read that like 3x before I understood what was wrong.  Very good catch!

Issue #372, opened 2021-09-21
Filling down data on copy paste broken


mdebrauw:
Selection should be filled down on pasting a single text / cell value. After MR #371 by @faimaklg this behaviour seems to be broken.

Expected behaviour: 

![CleanShot 2021-09-21 at 15 41 36](https://user-images.githubusercontent.com/101284/134181648-4f2b5588-c3a3-4629-aabc-72f05bdeab5c.gif)

Current behaviour:

![CleanShot 2021-09-21 at 15 37 09](https://user-images.githubusercontent.com/101284/134181117-8ddafc9d-1c7d-46c8-8c85-b7a2b5d16216.gif)

This seems to be related to changes in the [pasteDate](https://github.com/TonyGermaneri/canvas-datagrid/blob/92825b7fe5d2c101e4c51bffdb269e8890018c1f/lib/events.js#L1451) event handler and [parseData](https://github.com/TonyGermaneri/canvas-datagrid/blob/92825b7fe5d2c101e4c51bffdb269e8890018c1f/lib/events.js#L1478) function.

PR #371, opened 2021-08-30
update copy and paste.

faimaklg:
now copy and paste is more like excel

PR #373, opened 2021-09-21
Fix filling down on filling down single values on paste.

mdebrauw:
Selection should fill down on pasting a single text / cell value.

Bring back if-statement checking for single values and update.

Fixes issue #372

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1515,47 +1515,57 @@ export default function (self) {
 
     // selected cell. This mimics Excel's paste functionality, and works
     // as a poor-man's fill-down.
-    var selections = [];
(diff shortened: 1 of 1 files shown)


TonyGermaneri:
Is there a conflict in the desired behavior or is this just a bug or are you not sure?  I'm not sure.

mdebrauw:
It's a bug.

If you look at the linked issue I think it is clear? Before the change #371 you could fill down a selection with one value (text or one other cell, that is no longer possible. If you try to fill down a selection it will just paste the value in the first cell of the selection.

TonyGermaneri:
Thanks for tracking it down and fixing it!

PR #375, opened 2021-09-29
Shrink keyboard selections

mdebrauw:
When using keyboard navigation to select cells the selection only grows when additional rows or columns are added. 

CleanShot 2021-09-29 at 14 18 50

This is not in line with Excel/Google sheets behaviour and provides for a sub-optimal user-experience.

This PR changes the selection logic to allow for selections to be more flexible so that they can shrink if arrow keys are moved back:

CleanShot 2021-09-29 at 14 31 20

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1230,8 +1230,8 @@ export default function (self) {
     var i,
       ev,
       adjacentCells = self.getAdjacentCells(),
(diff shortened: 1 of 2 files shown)


ndrsn:
Excellent! Thank you!

PR #378, opened 2021-10-12
Handle clipboard data from Excel and other HTML formatted clipboard data

mdebrauw:
This MR handles tables, rows and individual cell paste actions from Excel correctly. It also handles with non-table `text/html` snippets to have more predictable copy/paste behaviour.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -12,7 +12,13 @@ module.exports = function (config) {
     // list of files / patterns to load in the browser
     files: [
       {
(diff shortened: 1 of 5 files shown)


mdebrauw:
Pushing an update: 
- Moving `events.js` into its own directory and extraced a couple of functions into `util.js` file.
- Cleanup regex expressions
- Adding support for Google Sheets tables (which was "out-of-the-box" by changing the regex expressions :)

PR #379, opened 2021-10-15
Optionally allow keyboard interaction when mouse cursor is not inside the grid

mdebrauw:
When mouse cursor moves outside of the grid the grid loses focus (`grid.hasFocus` set to `false`) disabling keyboard interaction.

This PR is a small usability tweak to allow keyboard interaction even when mouse is not hovering over the grid. Focus can also be set programmatically (`grid.focus`) to allow user input irrespective of mouse cursor location.


ndrsn:
Super!

PR #380, opened 2021-10-15
Add bound index map to selectionchanged event data

mdebrauw:
The event `selectionchanged` data does not contain the bound-data row/column index. This can be problematic when your application relies on being able to reference selected cells when the grid has applied filters or changed column order.

This PR adds two keys to the event data: `boundRowIndexMap` and `boundColumnIndexMap`, these contain an dict that keeps track of view-data and bound-data row/column indices. 

I will raise a separate issue (and MR) for updating the `selectionchanged` event as I think all current keys of the event could be rationalised to simply providing a 2 dimensional array that contains values and coordinates (view and bound) of the cells in the selection. @TonyGermaneri do you have thoughts on this?

### DIFF
--- a/lib/intf.js
+++ b/lib/intf.js
@@ -194,14 +194,9 @@ export default function (self, ctor) {
     }
 
     for (const [viewRowIndex, row] of self.selections.entries()) {
-      if (!row) continue;
+      if (!row || row.length === 0) continue;
 
-      if (row.length === 0) {
-        selectedCells[viewRowIndex] = null;
-        continue;
-      }
-
-      selectedCells[viewRowIndex] = [];
+      const currentRowSelections = [];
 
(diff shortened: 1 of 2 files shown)


mdebrauw:
Thinking this over a little more, I believe a better approach is to:

- Add a method `getSelectedCells` this returns a 2D array of rows and selected cells on those rows
- The rows contain a cell-like objects having `value`, `header`, `boundRowIndex`, `viewRowIndex`, `boundColumnIndex`, `viewColumnIndex` attributes
- Add a `selectedCells` key to the `selectionchanged` event data containing the output of `getSelectedCells()`

As a next step I think we can  remove the other keys (except `selectionBounds`) from the event data as all of these are supported by the new `selectedCells` key. This would be a breaking change, so maybe do it some other time?

I've updated PR with these changes.

PR #386, opened 2021-11-04
Fix Fiddle resource path

slominskir:
The tutorial Fiddles currently do not work.  The path given to canvas-datagrid.js file is no longer valid.


ndrsn:
Wonderful, thank you!

PR #388, opened 2021-11-05
fix string sort when value is null/undefined

faimaklg:
fix: string sort when value is null/undefined
动画

### DIFF
--- a/lib/publicMethods.js
+++ b/lib/publicMethods.js
@@ -1367,22 +1367,18 @@ export default function (self) {
   self.sorters.string = function (columnName, direction) {
     var asc = direction === 'asc';
     return function (a, b) {
-      if (a[columnName] === undefined || a[columnName] === null) {
-        return 1;
-      }
-      if (b[columnName] === undefined || b[columnName] === null) {
-        return 0;
-      }
+      const aValue = a[columnName] || '';
+      const bValue = b[columnName] || '';
       if (asc) {
-        if (!a[columnName].localeCompare) {
+        if (!aValue.localeCompare) {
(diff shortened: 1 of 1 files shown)


ndrsn:
Hi @faimaklg, thank you for this. I tried out this new version with some sample data, but it seems sorting still does not behave properly?

See this, for example:

CleanShot 2021-11-18 at 01 56 23

Data is

```js
[
            ['foo'],
            [null],
            ['baz'],
            ['quux']
]
```

I would expect `foo` to be included in the sorting fun, but it's stuck at the top, like a sad puppy.

Can you confirm this behaviour, or am I doing something wrong?

faimaklg:
> Hi @faimaklg, thank you for this. I tried out this new version with some sample data, but it seems sorting still does not behave properly?
> 
> See this, for example:
> 
> CleanShot 2021-11-18 at 01 56 23
> 
> Data is
> 
> ```js
> [
>             ['foo'],
>             [null],
>             ['baz'],
>             ['quux']
> ]
> ```
> 
> I would expect `foo` to be included in the sorting fun, but it's stuck at the top, like a sad puppy.
> 
> Can you confirm this behaviour, or am I doing something wrong?

I'm not sure if you used the wrong branch or schema data

See this, for example:
```
 const schema = [{"name":0, "title":"id", "type": "string"}];
  const data = [
      ['foo'],
      [null],
      ['baz'],
      ['quux']
  ];
```
动画

ndrsn:
Oh boy, you're totally right — I was still on `master`... 🤦‍♂️ 

Apologies!

PR #389, opened 2021-11-05
fix: endless loop aused by set orderDirection/orderBy when addEventLi…

faimaklg:
fix: endless loop aused by set orderDirection/orderBy when addEventListener('beforesortcolumn')
```
grid.addEventListener('beforesortcolumn', (e) => {
  e.preventDefault();
  grid.orderBy = e.name;
  grid.orderDirection = e.direction;
  // implement server side ordering
  grid.refresh();
});
```

### DIFF
--- a/lib/intf.js
+++ b/lib/intf.js
@@ -1080,7 +1080,6 @@ export default function (self, ctor) {
         value = 'asc';
       }
       self.orderDirection = value;
-      self.order(self.orderBy, self.orderDirection);
     },
   });
   Object.defineProperty(self.intf, 'orderBy', {
@@ -1096,7 +1095,6 @@ export default function (self, ctor) {
         throw new Error('Cannot sort by unknown column name.');
       }
       self.orderBy = value;
-      self.order(self.orderBy, self.orderDirection);
     },
   });
   if (self.isComponent) {


ndrsn:
I have two issues with this PR:

- First of all, it's a breaking change with unclear implications: simply removing the call to `.order()` will inevitably break existing users' applications.
- Second, I'm not sure this PR fixes an actual problem. Yes, there's an infinite loop, but I believe it's more to do with the code above than anything wrong with how canvas-datagrid operates.

To expand on the second point: I must admit I'm not clear what your code sample above is intending to accomplish; setting `orderBy` and `orderDirection` _outside_ of the dispatched event will call the `order` method, which dispatches the `beforesortcolumn` event. Then re-setting `orderBy` or `orderDirection` will then again call `.order()`, causing the infinite loop. It's not clear to me why you are re-setting these attributes?

faimaklg:
Actually, I want to sort by server，and set the sort icon for the column.
But now i cant find a function to set the icon.

ndrsn:
> For the first, It's better for the user to call grid..order() function by himself, instead of automatically after set grid.orderBy or grid.orderDirection.

I don't necessarily disagree with you, but it does not feel to me like introducing a breaking change is warranted here (yet).

> For the second, I want to sort by server，and set the sort icon for the column. But now i cant find a function to set the icon.

With "sort by server" do you mean you're sorting the data _outside_ of canvas-datagrid (whether that's server side, or in the client before passing it to the grid) ? Then I think I understand a little better. 

In that case, wouldn't it be better to disable sorting in the grid itself altogether? From there, the only problem you would have is how to draw a sort icon in the header. There is an example in the documentation [showing how to draw on the canvas](https://canvas-datagrid.js.org/#drawing-on-the-canvas) — following this example, you could hook into the `afterrrendercell` event, and check if it's the header cell for whatever column you're sorting, and if so draw your own icon.

faimaklg:
it's a good idea for the second. I close the pr.

Issue #393, opened 2021-11-06
Position of scroll-bar with a frozen pane


david542542:
When there is a frozen pane, it seems like the scroll-bar positioning is off. It should only allow going up to the frozen pane, since the frozen area is not part of the scroll.

![image](https://user-images.githubusercontent.com/909771/140593336-83c0a398-d547-48a5-b57d-4364037924fb.png)

For example: https://gyazo.com/93c2baba5d2cdd924ce65fe9f81a8eab.

ndrsn:
Nice catch — that's sloppy, we should fix it. I'll see if I can look into it this weekend.

Issue #390, opened 2021-11-06
Move multiple columns or Rows


david542542:
It would be great to have the ability to drag multiple columns (or Rows) when multiple are highlighted.

![image](https://user-images.githubusercontent.com/909771/140593205-253b0fca-97f1-440b-ad7d-399ebcccfa82.png)

Here is an example: example: https://gyazo.com/d32dfc66affb9b5de97fff1cbdb39a8b.

ndrsn:
I agree, this fits with our general aim to be more Excel/Sheets-like. @TonyGermaneri, any opinions on this?

hiro00125:
Hello
Can I work with you?

hiro00125:
I've just found out this issue on Upwork.
I'd like to help you.

david542542:
@hiro00125 once you delete all your comments, you can email me at deemarklit@gmail.com and we can follow up there regarding potential work.

PR #401, opened 2021-11-19
Fix moving multiple columns or rows (#393, #390)

xianzhi3:
Fixes issue #.

- fix Move multiple columns or Rows
- 
-

### DIFF
--- a/lib/draw.js
+++ b/lib/draw.js
@@ -122,15 +122,15 @@ export default function (self) {
     x = x + ml;
     y = y + mt;
     if (self.orderDirection === 'asc') {
-      self.ctx.moveTo(x, y);
-      self.ctx.lineTo(x + aw, y);
-      self.ctx.lineTo(x + aw * 0.5, y + ah);
-      self.ctx.moveTo(x, y);
-    } else {
       self.ctx.lineTo(x, y + ah);
       self.ctx.lineTo(x + aw, y + ah);
       self.ctx.lineTo(x + aw * 0.5, y);
       self.ctx.lineTo(x, y + ah);
+    } else {
+      self.ctx.moveTo(x, y);
+      self.ctx.lineTo(x + aw, y);
+      self.ctx.lineTo(x + aw * 0.5, y + ah);
+      self.ctx.moveTo(x, y);
     }
     self.ctx.stroke();
     self.ctx.fill();
--- a/lib/draw.js
+++ b/lib/draw.js
@@ -1514,6 +1514,10 @@ export default function (self) {
       self.ctx.lineWidth = self.style.reorderMarkerBorderWidth;
       self.ctx.strokeStyle = self.style.reorderMarkerBorderColor;
       if (self.dragMode === 'row-reorder') {
+        for (var k = 0; k < self.selections.length; k++) {
+          if (!self.selections[k] || k == self.reorderObject.rowIndex) continue;
+          b.height += self.getRowHeight(k);
+        }
         b.width = w;
         b.x = 0;
         m.width = w;
@@ -1536,6 +1540,12 @@ export default function (self) {
           );
         }
       } else if (self.dragMode === 'column-reorder' && self.reorderObject) {
+        var selectedColumns = self.selections[0];
+        for (var k = 1; k < selectedColumns.length; k++) {
+          b.width += self.getColumnWidth(
+            self.orders.columns[selectedColumns[k]],
+          );
+        }
         b.height = h;
         b.y = 0;
         m.height = h;
--- a/lib/draw.js
+++ b/lib/draw.js
@@ -601,15 +601,28 @@ export default function (self) {
         m = self.style.scrollBarBoxMargin * 2;
       self.ctx.strokeStyle = self.style.scrollBarBorderColor;
       self.ctx.lineWidth = self.style.scrollBarBorderWidth;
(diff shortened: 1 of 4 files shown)


ndrsn:
CleanShot 2021-11-23 at 13 35 07

Seems to be a bug, where it copies the first column `VIOLATION LAST MODIFIED DATE`, replacing `VIOLATION DATE` ? I'm  Command/Ctrl clicking the column headers

xianzhi3:
ok, let me check the issue again and fix it

xianzhi3:
@ndrsn I fixed the bug. you can test at the link I sent.

lib/draw.js
```diff
@@ -601,15 +601,28 @@ export default function (self) {
         m = self.style.scrollBarBoxMargin * 2;
       self.ctx.strokeStyle = self.style.scrollBarBorderColor;
       self.ctx.lineWidth = self.style.scrollBarBorderWidth;
-      en.horizontalBox.x =
-        rowHeaderCellWidth +
-        self.style.scrollBarBoxMargin +
-        (en.horizontalBar.width - self.scrollBox.scrollBoxWidth) *
-          (self.scrollBox.scrollLeft / self.scrollBox.scrollWidth);
+      if (self.frozenColumn > 0)
```
ndrsn, line 604: Technically this _is_ just one statement (as with the `else` block), but because it's such a long one, maybe it's better to add braces `{ ... }` around these `if` and `else` blocks for clarity? Or did prettier format it like this?

lib/events/index.js
```diff
@@ -254,7 +254,11 @@ export default function (self) {
     setScrollBoxSize();
     self.scrollBox.scrollWidth = dataWidth - self.scrollBox.width;
     self.scrollBox.scrollHeight = dataHeight - self.scrollBox.height;
-    self.scrollBox.widthBoxRatio = self.scrollBox.width / dataWidth;
+    if (self.frozenColumn > 0)
```
ndrsn, line 257: Same comment as above; I'd prefer braces around anything but the simplest of `if` statements. For instance:

```js
if (foo === false) return;
```

is fine. But

```js
if (foo === false) bar = quux + bar === 123 ? getSomething() : getSomethingElse();
else return doSomethingAndReturnIt();
```
is (IMHO) already a bit risky. What do you think?

lib/events/index.js
```diff
@@ -546,6 +554,11 @@ export default function (self) {
             }
           } else if (dragBounds.top !== -1) {
             self.selectArea(sBounds, true);
+            if (
+              sBounds.left == -1 &&
+              Math.abs(sBounds.top - sBounds.bottom) > 0
```
ndrsn, line 559: Not sure I completely understand what's going on here: `Math.abs` will turn `sBounds.top - sBounds.bottom` into a positive number if it's negative, so this expression will either be a positive number or zero; so I believe this expression is equivalent to checking whether `sBounds.top` and `sBounds.bottom` are different? As in:

```js
Math.abs(sBounds.top - sBounds.bottom) > 0
// is equivalent to:
sBounds.top !== sBounds.bottom
```

Is that correct? If so, I think this is maybe clearer? What do you think?

lib/events/index.js
```diff
@@ -641,11 +654,11 @@ export default function (self) {
           checkSelectionChange();
           return;
         }
-        if (self.attributes.columnHeaderClickBehavior === 'select') {
+        /*if (self.attributes.columnHeaderClickBehavior === 'select') {
```
ndrsn, line 657: Can we get rid of this comment entirely? Or do we need it later?

lib/events/index.js
```diff
@@ -996,7 +1052,12 @@ export default function (self) {
     ) {
       return;
     }
-    var pos = self.getLayerPos(e);
+    var pos = self.getLayerPos(e),
+      bm = self.style.gridBorderCollapse === 'collapse' ? 1 : 2,
```
ndrsn, line 1056: What is `bm` here? Is there a more descriptive name possible?

lib/publicMethods.js
```diff
@@ -379,6 +379,37 @@ export default function (self) {
...
+   * Returns true if indices of columns next to the selected columnIndex is selected on every row.
+   * @memberof canvasDatagrid
+   * @name isMultiColumnsSelected
+   * @method
+   * @param {number} columnIndex The column index to check.
+   */
+  self.isMultiColumnsSelected = function (columnIndex) {
+    var multiColIsSelected = true;
+    self.viewData.forEach(function (row, rowIndex) {
+      var columnIndics = self.selections[rowIndex];
```
ndrsn, line 392: Rename `columnIndics -> columnIndices` ?

ndrsn (changes requested):
Nice! We're very close! Just a few small changes requested

lib/events/index.js
xianzhi3: yes, that is correct

lib/events/index.js
xianzhi3: ok let me get rid of the comment

lib/events/index.js
xianzhi3: http://prntscr.com/20sd870
I referred original code.

lib/events/index.js
xianzhi3: ok, that's great

PR #402, opened 2021-11-24
fix: cant copy when column name or cell value is number

faimaklg:
cant copy when column name or cell value is number

example data
```
const schema = [{"name": 0, "title":"id", "type": "number"}, {"name": 1, "title":"name", "type": "string"}];
const data = [
    [1, 'foo'],
    [2, null],
];
```


ndrsn:
Thanks you! Just released this in v0.3.19.

PR #441, opened 2021-11-28
WIP: Migrate docs to docusaurus (alpha version)

ndrsn:
_Note: This is a work in progress and should not be merged until the tutorials have been included in the docs, and we've checked all the links still work._

This PR migrates the documentation to use [Docusaurus](https://docusaurus.io/), hopefully making maintenance and changing the documentation easier and more straightforward. It splits up the various sections in `canvasDatagrid.md` into separate Markdown files (see them in `docs/docs/`), and adds a `docs/jsdoc-to-md.js` file that converts the JSDoc annotations to Markdown files (located in `docs/reference/` upon generation) using Handlebars templates. A `README.md` file in `docs/` has some more detail.

I think I covered most/all of the sections in `canvasDatagrid.md`, but please let me know if I missed anything. Also, the generated files in `reference/` aren't perfect: I'm not sure they're terribly clear, and if the layout maybe can be improved there. For now, the structure/sections are mostly kept as-is, but I can imagine we'll eventually want to restructure a bit. Any comments or ideas on that are welcome.

What's missing from the docs still is the tutorials: I just haven't gotten around to it yet, and wanted to check in that this the right approach first. The tutorials would need a JS Fiddle or CodeSandbox or similar button as we have now. Ideally, I see each tutorial as a separate markdown file in `docs/tutorials/`, each with one or more code blocks which automagically are turned into JSFiddles upon page load. Not sure how to get there yet, but that would be the dream, I think.

There's a preview of the documentation here: https://likablestimulatingdatamining.ndrsn.repl.co/


faimaklg:
good job

Issue #398, opened 2021-11-08
Possible to create a 'button' within a cell?


david542542:
In Excel and Google Sheets a 'button' is often used for a context menu related to a heading, for example:

![image](https://user-images.githubusercontent.com/909771/140803689-adb63c84-7af1-421e-ac99-a6f3a4f123a8.png)

Does canvas-datagrid currently support this, or is it something that would need to be added as a new feature? If it's something that's already supported, could we add an example of it to the [Tutorials](https://canvas-datagrid.js.org/#tutorials) section?

ndrsn:
It's not supported out of the box, but you can hook into the `afterrendercell` event to draw your own 'button' on top of the canvas. [Here's a crude example](https://replit.com/@ndrsn/ThreadbareReadyPagerecognition#script.js) (it's a box not a button, but you get the idea):

<img width="848" alt="CleanShot 2021-11-12 at 14 17 44@2x" src="https://user-images.githubusercontent.com/950979/141473269-d15e6243-a73f-4f84-88f8-254fbd8b4f9b.png">

```js
grid.addEventListener('afterrendercell', (event) => {
  if (event.cell.rowIndex === -1 && event.cell.columnIndex > -1) {
    drawButton(event.ctx, event.cell);
  }
});
```

Obviously, you'd want to make it actually look like a button, render differently on hover or active states, but what this looks like depends a little bit on what the rest of your application looks like — I don't think canvas-datagrid should be too opinionated in terms of UI look and feel, or you lose flexibility when integrating.

Alternatively, you could use HTML buttons. I have the same issue: I also want to show filter buttons like Excel, and canvas drawing seems cumbersome, so this is the route I've wanted to investigate, but haven't gotten around to yet.

hiro00125:
Hello
Can I work with you?
I've just found out this issue on Upwork.

david542542:
@hiro00125 delete this comment (and the other one) and apply directly, please.

david542542:
@xianzhi3 here are two more examples of where a button could be used: https://gyazo.com/905512e26cb23f9458d3a2fc66532fcb. In other words, it would be great to be able to arbitrarily position and style the buttons (ignore what they produce after clicked).

david542542:
[...]

PR #442, opened 2021-12-03
fix Possible to create a button within a cell?

xianzhi3:
Fixes issue #.

- 
- 
-


xianzhi3:
@ndrsn I added new features for this issue #398 . You can have a test at the link below.
https://90zfd.csb.app/

lib/button.js
```diff
@@ -0,0 +1,326 @@
...
+
+    intf.dispose = function () {
+      if (wrapper.parentNode) {
+        wrapper.parentNode.removeChild(wrapper);
+      }
+    };
+    return intf;
+  }
+
+  function toogleButtonMenu() {
```
david542542, line 99: typo?

lib/contextMenu.js
```diff
@@ -173,6 +173,33 @@ export default function (self) {
         self.parentDOMNode.appendChild(downArrow);
       }
     }
+    function fade(element) {
+      var op = 1;
```
ndrsn, line 177: Where possible, please use descriptive variable names — for instance, I think `op -> opacity` is more clear?

lib/contextMenu.js
```diff
@@ -173,6 +173,33 @@ export default function (self) {
...
+      var timer = setInterval(function () {
```
ndrsn, line 178: If `timer` is not used anywhere, do we still need it?

lib/contextMenu.js
```diff
@@ -173,6 +173,33 @@ export default function (self) {
...
+        }
+        element.style.opacity = op;
+        element.style.filter = 'alpha(opacity=' + op * 100 + ')';
+        op -= op * 0.1;
+      }, self.attributes.animationHideContextMenu * 0.1);
+    }
+    function unfade(element) {
+      var op = 0.1;
+      element.style.display = 'block';
+      var timer = setInterval(function () {
```
ndrsn, line 194: Same

lib/contextMenu.js
```diff
@@ -173,6 +173,33 @@ export default function (self) {
...
+            element.parentNode.removeChild(element);
+          }
+        }
+        element.style.opacity = op;
+        element.style.filter = 'alpha(opacity=' + op * 100 + ')';
+        op -= op * 0.1;
+      }, self.attributes.animationHideContextMenu * 0.1);
+    }
+    function unfade(element) {
+      var op = 0.1;
```
ndrsn, line 192: Same (re: opacity)

lib/contextMenu.js
```diff
@@ -664,23 +691,28 @@ export default function (self) {
       autoCompleteContext = undefined;
     }
   };
-  self.disposeContextMenu = function () {
+  self.disposeContextMenu = function (e) {
```
ndrsn, line 694: Is `e` an event? If so please rename to `event` for legibility.

lib/contextMenu.js
```diff
@@ -722,8 +754,15 @@ export default function (self) {
         contextPosition.width =
           self.width - self.style.mobileContextMenuMargin * 2 + 'px';
       }
+      if (e.type == 'mousedown') {
+        contextPosition.top += 10;
```
ndrsn, line 758: It's not clear to me exactly what this `10` represents — can we assign it to a variable, so it has a name? It seems to be an offset of some kind; would it make sense to make it a configurable style option?

lib/defaults.js
```diff
@@ -16,6 +16,8 @@ export default function (self) {
       ['allowRowResize', true],
       ['allowRowResizeFromCell', false],
       ['allowSorting', true],
+      ['animationShowContextMenu', 50],
```
ndrsn, line 19: `animationShowContextMenu -> animationDurationShowContextMenu`
`animationHideContextMenu -> animationDurationHideContextMenu`

lib/events/index.js
```diff
@@ -1556,7 +1605,10 @@ export default function (self) {
       self.currentCell.style === 'cornerCell'
     ) {
       self.autosize();
-    } else if (['cell', 'activeCell'].indexOf(self.currentCell.style) !== -1) {
+    } else if (
+      ['cell', 'activeCell'].indexOf(self.currentCell.style) !== -1 &&
```
ndrsn, line 1609: Prefer using `.includes()` over `indexOf() !== -1`

lib/button.js
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  function createButtonMenu(pos, items) {
+    var buttonMenu = document.createElement('div'),
+      selectedIndex = -1,
+      intf = {},
+      rect;
+
+    function createItems() {
+      items.forEach(function (item) {
```
ndrsn, line 131: For all applicable occurrences in this PR (where it makes sense, that is), would it be possible to replace `.forEach` with a for-of loop? For example,

```js
for (const el of array) {
  // do something
}
```

They are [faster than forEach](https://coderwall.com/p/kvzbpa/don-t-use-array-foreach-use-for-instead), and I find the syntactic overhead of a function distracting and unnecessary.

lib/button.js
```diff
@@ -0,0 +1,326 @@
+/*jslint browser: true, unparam: true, todo: true*/
+/*globals define: true, MutationObserver: false, requestAnimationFrame: false, performance: false, btoa: false, Event: false*/
+'use strict';
+
+export default function (self) {
+  var zIndexTop;
```
ndrsn, line 6: For future PRs, would you be willing to use `const` or `let` instead? Not a huge deal, because the current code base is still littered with `var`, but we're trying to migrate away, and smaller scoping would make bug-fixing and refactoring in the future easier.

lib/button.js
```diff
@@ -0,0 +1,326 @@
...
+      loc.x = pos.left - s.left;
+      loc.y = pos.top - s.top;
+      loc.height = 0;
+      zIndexTop += 1;
+      buttonMenu.style.position = 'absolute';
+      buttonMenu.style.zIndex = zIndexTop;
+      buttonMenu.style.left = loc.x + 'px';
+      buttonMenu.style.top = loc.y + 'px';
+      document.body.appendChild(buttonMenu);
+      rect = buttonMenu.getBoundingClientRect(); // TODO: fix
```
ndrsn, line 192: Fix what? 😊

lib/button.js
```diff
@@ -0,0 +1,326 @@
...
+        buttonMenu.parentNode.removeChild(buttonMenu);
+      }
+    };
+
+    Object.defineProperty(intf, 'selectedIndex', {
+      get: function get() {
+        return selectedIndex;
+      },
+      set: function set(value) {
+        if (typeof value !== 'number' || isNaN(value || !isFinite(value))) {
```
ndrsn, line 246: Should `isNaN(value || !isFinite(value))` actually be `isNaN(value) || !isFinite(value)` ?

ndrsn (changes requested):
Great work, @xianzhi3! I have a few comments, but nothing too big. Overall really well-written, clear code 👍 

Something else though, it seems as though the PR includes commits from previous PRs? It makes it harder to review; I'm not sure what reason is, but would you see if for future PRs you can only include commits that are relevant to the issue at hand? Thanks!

xianzhi3:
Ok, will do. thanks

lib/contextMenu.js
xianzhi3: timer is used for clearInterval

lib/button.js
xianzhi3: ok

lib/contextMenu.js
ndrsn: But is `clearInterval` used? I don't see it.

xianzhi3: Screenshot_2

lib/button.js
ndrsn: The comment's gone now, but was there a fix needed, or was there another reason for this comment?

xianzhi3: sorry, it was my mistake

ndrsn:
> Ok, will do. thanks

Great, thanks! Just a couple questions still, but looks good to merge otherwise 👍

lib/contextMenu.js
ndrsn: Oh, wow — that's embarrassing 🤦 

Apologies!

ndrsn:
Great! If you can fix the merge conflict with `lib/events/index.js`, I'll merge this.

luckyfanghe:
I can't see the Resolve conflicts. Can you check the permission?

ndrsn:
Hmm, not sure what's going on... this is what I see:

CleanShot 2021-12-04 at 11 32 21@2x

ndrsn:
I think the "Resolve conflicts" button is only available to maintainers

luckyfanghe:
I was able to see the button on the PR I did for moving multiple columns before. I keep the message
image

ndrsn:
Not sure what's different compared to previous PR, but that means you would need to resolve the merge conflicts locally and push the commit, instead of doing it through the GitHub web interface.

The fix seemed easy enough, so I resolved it myself. Will merge. Thank you for all the work, @xianzhi3 (and @luckyfanghe?)!

Issue #412, opened 2021-11-25
Ability to clear data in a cell selection


david542542:
It would be nice when there is a cell selection to be able to press `Delete` and the data to be removed. For example: https://gyazo.com/6bc63aeceaa1331217f90154e101e730

PR #446, opened 2021-12-06
Add `afterdelete` event and `deleteSelectedData()` public method

mdebrauw:
Natively support clearing selected data pressing Delete/Backspace. Clearing selected data fires a `afterdelete` event containing affected cells (an array of tuples in the shape of `[rowIndex, columnIndex, boundRowIndex, boundColumnIndex]`).

This addresses #412.

See https://ynoqp.csb.app/ for interactive demo.

### DIFF
--- a/lib/docs.js
+++ b/lib/docs.js
@@ -461,6 +461,14 @@
  * @param {object} e.boundColumnIndexMap Mapping of view column order to bound column order.
  */
 /**
+ * Fires when the selection is cleared by pressing backspace or delete key.
+ * @event
+ * @name canvasDatagrid#selectioncleared
+ * @param {object} e Event object.
+ * @param {object} e.ctx Canvas context.
+ * @param {array} e.cells Cells affected by the clear action. Each item in the array is a tuple of [rowIndex, columnIndex, boundRowIndex, boundColumnIndex].
+ */
+/**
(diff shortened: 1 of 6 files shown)


ndrsn:
Nice! Thanks!

ndrsn:
Doesn't seem to clear until after the mouse moves:

CleanShot 2021-12-06 at 21 41 27

TonyGermaneri:
Adding a draw event to the delete method might cause a problem with API based deletes.  Maybe a debounced draw after 10ms timeout?   That way if someone spams the delete method it won't try and redraw between each call.  Might not even be a problem though.

mdebrauw:
Good catch, strange that I missed that. I've added a call to `self.draw()` to the public method `deleteSelectedData()`. 

I guess it would be up to the API user (i.e. someone calling  `deleteSelectedData()`) to make sure not to 'spam' it, right? Either they would debounce calls to it or use `dontDraw` parameter to avoid redraws.

@TonyGermaneri maybe I missed other ways to do API based deletes, or were you referring to the newly introduced `deleteSelectedData()`?

mdebrauw:
The `cut` event handler suffered from same issue, grid was only redrawn after mouse move or keyboard action. Fixed this.

PR #447, opened 2021-12-06
Fire event when resizing column by double clicking

mdebrauw:
Double clicking a column border autoresizes the column width (to fit widest cell) but does not emit a `resizecolumn` event. 

This PR fixes this.


ndrsn:
👍

PR #450, opened 2021-12-09
Refactor `cut` and `copy` event handlers

mdebrauw:
The cut event handler currently dispatches a `copy` and `cut` event. The copy is fired by reuse of the `copy` event handler and the data is removed from the grid before the `cut` event is dispatched. This is not in line with the [W3C Clipboard API](https://w3c.github.io/clipboard-apis/#clipboard-event-cut) which reads:

> In an editable context, if the event is not canceled the action will place the currently selected data on the system clipboard and remove the selection from the document. The cut event fires before the selected data is removed. When the cut operation is completed, the selection is collapsed.

This PR refactors the `cut` event handler to:

- Move selected-cell-copy-logic to a utility function so it can be used in both `cut` and `copy` event handler.
- Copy selected data to the clipboard but not dispatch a `copy` event.
- (Re)-Dispatch  `cut` event before anything happens.
- Clear selected data and then dispatch a custom`aftercut` event with an array of affected cells (usable in calling application).
- Add/enable tests for neat and untidy copy actions.

In #383 I added `affectedCells` to the `cut` event, this PR breaks this by introducing an `aftercut` event with those cells. I would think impact of this is very limited.

Here's a demo: https://3nvso.csb.app/

### DIFF
--- a/lib/events/index.js
+++ b/lib/events/index.js
@@ -1842,35 +1842,22 @@ export default function (self) {
     });
   };
   self.cut = function (e) {
-    self.copy(e);
-    const schema = self.getSchema();
-
-    var affectedCells = [];
-    for (const [rowIndex, row] of self.selections.entries()) {
-      if (!row) continue;
-
-      const boundRowIndex = self.getBoundRowIndexFromViewRowIndex(rowIndex);
-
-      for (const columnIndex of row) {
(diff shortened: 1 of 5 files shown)


ndrsn:
Very nice, thank you!

PR #443, opened 2021-12-04
Support grouping columns

hangxingliu:
For issue #397

This Pull Request is for supporting column group, and this is the demo:

<https://user-images.githubusercontent.com/4303130/144727083-e19df740-5173-432d-9d2a-29fc1bf0661a.mp4>

And I will create another Pull Request about the row grouping in the next week.

PR #453, opened 2021-12-18
Allowing a tree grouping along columns

xianzhi3:
Fixes issue #.

- 
- 
-


david542542:
@xianzhi3 + @ndrsn looks very good to me: https://90zfd.csb.app/. Nice work.

ndrsn:
This seems like a bug?

CleanShot 2021-12-20 at 10 14 47

xianzhi3:
let me check

xianzhi3:
it looks like the issue is not one that is related with the tree group. When you resize the columns, it is only scrolled right.

david542542:
@xianzhi3 that's wrong. Here is a video. Please fix all issues mentioned in the video: https://gyazo.com/9547e615a7ead1e46a0c34388fee8355

- Dragging right should only change the right-side width of an item
- Dragging left should only change the left-side width of a grid

However! I think this may already be working properly, and there is a bug with the scroll-bar, so I think if you examine the scroll-bar positioning (and it's calculation of where to put it) that is where the issue comes from.

david542542:
Yes, @xianzhi3 + @ndrsn I think my suspicion is correct. It is a scroll-bar positioning issue. Nonetheless, please fix this:

https://gyazo.com/cc76d70142c889b6003b153779b66729

xianzhi3:
ok, let me check scroll-bar part

xianzhi3:
I fixed the bug. You can have a test at the link.
https://90zfd.csb.app/

ndrsn:
@xianzhi3 , apologies for getting back to you so late on this. Thanks for all the work and the bugfix! I have noticed something else, though:

CleanShot 2022-01-05 at 16 37 03

When collapsing, I wouldn't expect the value immediately below the group handle to disappear — or is this as intended?

Another issue: when typing a value below a column tree group, it seems to not render, although when editing the value seems to still exist on that cell:

CleanShot 2022-01-05 at 16 40 43

Finally, when typing next to the tree, it seems to delete cell values. Not sure how this works with pre-existing cell values, though...

CleanShot 2022-01-05 at 16 54 15

lib/intf.js
```diff
@@ -71,6 +71,17 @@ export default function (self, ctor) {
     columnIndex: -1,
     rowIndex: -1,
   };
+  self.celltree = {
```
ndrsn, line 74: For consistency I think `cellTree` would be better?

lib/draw.js
```diff
@@ -164,6 +164,54 @@ export default function (self) {
     self.ctx.fill();
     return ml + aw + mr;
   }
+  function drawCellTreeIcon(cell, tree, rowTree) {
+    const pc = rowTree ? tree.parentCount : 0;
```
ndrsn, line 168: Can we write out variable names like `pc/rc/mt/mr/ml` etc to something more descriptive?

No hard rule that _all_ need to be written out (I mean, stuff like `x` and `y` are obvious, as are counters like `i`), but while it's a little more annoying to write the code, writing the more semantic variables out makes it _a lot_ easier to read and debug afterwards

ndrsn (changes requested):
Love the work in general (!), it's nicely structured and clear, but have some minor comments/requests.

Also, would it be possible to add tests for this functionality, similar to how @hangxingliu has done in #443 ? This will make refactoring down the road a whole lot easier/safer :)

I'm not sure how exactly that would look but can help with pointers if you need assistance.

xianzhi3:
"When collapsing, I wouldn't expect the value immediately below the group handle to disappear — or is this as intended?"
it is collapsed with options I added because the table is pivot one.

david542542:
@xianzhi3 no, he means this: https://gyazo.com/6162bf0bf7045a44d4d03fe27412b1d2

xianzhi3:
@ndrsn I fixed the bugs above and pushed the changes

xianzhi3:
https://user-images.githubusercontent.com/78894564/148524779-688faeb4-228e-481d-9bf1-3583946aeedd.mp4

Here is a video for the fix

ndrsn:
Awesome work, @xianzhi3! Looks really good 👍 

One last thing I'm still seeing is the disappearing value below the column group handles:

CleanShot 2022-01-07 at 11 50 39

Is this how Excel does it as well? It seems counterintuitive.

david542542:
Hi @ndrsn this should be the expected behavior for a pivot table. Here would be an example from Excel: https://gyazo.com/7cd2a84cecc9c506ff8fc2bfa2a289c9.

I suppose the best way to describe it would be to 'hide details' and summarize to a higher level. For example, if I have `Year > Month` collapsing to the Year would show the value at the year-level and hide the details of the month-level.

ndrsn:
@david542542 correct me if I'm wrong, but I think the difference between this implementation and what I see in your Excel example, is that the column immediately below the plus icon remains the same, whereas in Excel the column is replaced with the one at the end of the column group.

This is hard to describe in text, so take 0:55 in your video, for example. Once the group is collapsed, the values in column P are replaced with those of column W (`AUD` and `3.92` disappear).

Before collapsing:
CleanShot 2022-01-08 at 15 32 58@2x

After collapsing:
CleanShot 2022-01-08 at 15 33 50@2x

 In the implementation in this PR, only the value under the icon disappears, but the column values in other rows remain?

david542542:
@ndrsn ah yes I see. let me write up an example for @xianzhi3 to fix.

david542542:
@xianzhi3 I have added a sample Excel file and here's a video of how I want you to test: https://gyazo.com/902d4f84a14b3fdf60e610593d009cba. It's going to be tedious, but for the pivot table (you can hard-code the data that's fine) I want you to (1) collapse all-columns and make sure the values are the exact same as excel; then (2) collapse all-rows and make sure they're the same as excel; and then (3) collapse both and make sure they're the same. The testing will be tedious, but yes this is what you'll need to do in your video to confirm it's working properly.

[TestPivot.xlsx](https://github.com/TonyGermaneri/canvas-datagrid/files/7836516/TestPivot.xlsx)

xianzhi3:
I pushed the changes to the link below.
https://90zfd.csb.app/
Here is a video for that.

https://user-images.githubusercontent.com/78894564/148806913-0f1160bc-8b18-4394-9750-6336fa5b71cb.mp4

ndrsn:
Excellent, @xianzhi3, thanks so much! This is really great work. Merging!

🎉 🎈 💥

david542542:
@xianzhi3 also, let's please fix the context menu now -- it seems to be missing the various Expand/Collapse node options: https://gyazo.com/980b51159c5464a295f7b3076b1ecad9

PR #466, opened 2022-01-10
Fix jsDoc for document site

hangxingliu:
@ndrsn Urgent patch for <https://canvas-datagrid.js.org/>

Sorry, I missed description text in two functions I wrote. They causes the tutorials feature of current document site down, because the site generator needs generate HTML by `description.innerHTML = marked(formatDocString(i.description));`.


ndrsn:
Great, thank you for catching this!

Issue #469, opened 2022-01-15
The behavior of the menu item for hiding column becomes incorrect after the user reordered columns


hangxingliu:
### Steps to reproduce the problem.

https://user-images.githubusercontent.com/4303130/149609204-82f15b6a-49e6-4480-a6d4-75ac3d8f28f5.mp4

It is caused by `self.getSchema()[e.cell.columnIndex].hidden = true;` in `contextMenu.js`.

And I will fix this bug later

hangxingliu:
https://github.com/hangxingliu/canvas-datagrid/commit/36d8714951c468992f3667d99136ece9f624463c

PR #470, opened 2022-01-15
Fix the behavior of the conext menu item used to hide column

hangxingliu:
Fixes issue #469 

And add test case for it.


ndrsn:
👍

PR #482, opened 2022-02-05
Add vue setup instructions

boukeversteegh:
Hi! Fantastic library!

Although it's so simple, I still didn't immediately get it setup correctly in my Vue Typescript project. The reason is it wasn't entirely clear if I could use the `canvas-datagrid` element right away, or if I had to change any webpack configuration. I've never worked with web-components before.

Apparently, I just had to import the dependency in `main.ts` (or `.js`), and that's it.
I also needed to google for where to put the configuration for ignoredElements.

To make it even easier to get started with Vue, I thought to add these notes to the Vue demo page.

This is what it looks like:

https://jsfiddle.net/vs9k0nau/

### DIFF
--- a/tutorials/vueExample.html
+++ b/tutorials/vueExample.html
@@ -6,17 +6,30 @@
     </head>
     <body>
         <div id="app">
-            <!-- to get canvas-datagrid working in vue make sure you use the .prop attribute suffix -->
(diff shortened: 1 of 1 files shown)


ndrsn:
Hoi Bouke, thanks for this!

PR #490, opened 2022-02-18
add feature to move a cell or range of cells

xianzhi3:
Here is a link for test
https://90zfd.csb.app/413.html

### DIFF
--- a/lib/events/index.js
+++ b/lib/events/index.js
@@ -1219,7 +1219,8 @@ export default function (self) {
   self.stopDragMove = function (e) {
     document.body.removeEventListener('mousemove', self.dragMove, false);
     document.body.removeEventListener('mouseup', self.stopDragMove, false);
-    var b = self.getSelectionBounds();
+    var b = self.getSelectionBounds(),
+      s = self.getSchema();
     if (
       self.dispatchEvent('endmove', { NativeEvent: e, cell: self.currentCell })
     ) {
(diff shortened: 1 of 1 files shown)


ndrsn (changes requested):
I'm not sure this PR is fully complete, yet? Is the intention here to _only_ modify the viewed data, which may change at any time, not the underlying original data?

I would imagine that when you're moving a bunch of data around in a spreadsheet, the user would want the original data changed as well, not just what's painted on the screen.

Would you not need to modify  `originalData` then as well? And, preferably, also emit the moved cells in an (`endmove`?) event?

xianzhi3:
> I'm not sure this PR is fully complete, yet? Is the intention here to _only_ modify the viewed data, which may change at any time, not the underlying original data?
> 
> I would imagine that when you're moving a bunch of data around in a spreadsheet, the user would want the original data changed as well, not just what's painted on the screen.
> 
> Would you not need to modify `originalData` then as well? And, preferably, also emit the moved cells in an (`endmove`?) event?

xianzhi3:
Ok

xianzhi3:
Ok, let me check

hangxingliu:
Hi @ndrsn @xianzhi3 , I think this PR can be closed. I am refactoring the selection feature of this grid. So the code in this PR will be replaced in the next PR. I am going to sleep now and I will create a new PR in the next week

ndrsn:
Alright, thanks for the update!

PR #491, opened 2022-02-18
Dispatch column hide/unhide events

mdebrauw:
Dispatching events on hiding/unhiding columns is useful te be able to store grid state externally, (e.g. in settings or user profile). 

I've now made it so that an event is dispatched for each individual column being hidden even when a group of columns is being hidden. User can handle all individual events or debounce and, on last event, retrieve grid schema to figure out which columns are hidden or not.

Also moved `unhideColumns` to `publicMethods.js`, this seemed appropriate as all the other hide/unhide methods are located there.

Similar approach could be implemented for hiding / unhiding rows, will do as separate PR if agreed on this approach for columns.

### DIFF
--- a/lib/docs.js
+++ b/lib/docs.js
@@ -974,6 +974,22 @@
  * @param {canvasDatagrid.cell} e.cell The cell being moved out of.
  */
 /**
+ * Fires column(s) are being hidden
+ * @event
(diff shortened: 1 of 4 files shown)


ndrsn:
Very nice, thank you!

PR #483, opened 2022-02-07
change frozen style like google spreadsheet

xianzhi3:
Hi, @ndrsn I changed the frozen style like google spreadsheet.
You can have a test at the link below.
https://90zfd.csb.app/demo.html

PR #494, opened 2022-02-22
Fix the bug that the canvas being cleared under unusual refresh flow

hangxingliu:
The PR is used for fixing the bug in the PR #483 

Here is a explaining of it:

1. Setting the size properties (`width`/`height`) to the canvas element can causes the canvas and its state be cleared even if the size is the same.
2. So the grid loss all rendered content if we don't draw again after a function call to `self.resize()`

Even though this PR fixes it. But I still recommend using the code `self.resize(true)` to instead of the following code:

``` javascript
self.draw(true);
self.resize();

// or:
self.resize();
self.draw(true);
```

Because it can reduces unnecessary function calls. (Because `self.draw` may also call the `self.resize` if the `checkScrollHeight` is true. https://github.com/TonyGermaneri/canvas-datagrid/blob/master/lib/draw.js#L2890-L2892)

### DIFF
--- a/lib/events/index.js
+++ b/lib/events/index.js
@@ -386,9 +386,17 @@ export default function (self) {
     );
     // set canvas drawing related items
     if (!self.isChildGrid) {
-      self.canvas.width = self.width * ratio;
-      self.canvas.height = self.height * ratio;
-      self.ctx.scale(ratio, ratio);
+      const newWidth = self.width * ratio;
+      const newHeight = self.height * ratio;
+      // We need to check is settings size to canvas necessary,
+      // because settings the canvas'size will cause the canvas and its state be cleared
+      // even if the size is the same.
+      // Notes: Please don't call `self.resize()` without a subsequent call to `self.draw()`
+      if (self.canvas.width !== newWidth || self.canvas.height !== newHeight) {
+        self.canvas.width = newWidth;
+        self.canvas.height = newHeight;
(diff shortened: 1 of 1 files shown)


hangxingliu:
By the way, here is a trace script for internal function calls. I think it may be useful for the developers: <https://github.com/hangxingliu/canvas-datagrid-patch/blob/master/files/tutorials/hangxingliu/debugger.js>

You get something like this screenshot in your devTools if you use it:
Screenshot 2022-02-22 at 13 06 51

ndrsn:
> By the way, here is a trace script for internal function calls. I think it may be useful for the developers

I agree! Would you like to open a PR for this to merge it in?

hangxingliu:
> > By the way, here is a trace script for internal function calls. I think it may be useful for the developers
> 
> I agree! Would you like to open a PR for this to merge it in?

Hi @ndrsn ,I have two points about this debug script:

1. I don't know where can I put it into this repository. Because It is not for production environment, so it can't be packed into the distributed files.
2. It is a temporary solution for tracking. It injects code on the rumtime level, so it has some disadvantages. (Eg: it can't prints calls to closure function, and it can't know who call the function if a function is called asynchronized(`setTimeout`, `requestAnimationFrame`...)). I think that I can migrate an webpack injection for tracking from my other project to this project when I am free

Issue #435, opened 2021-11-26
Select-All implementation


david542542:
Select-all (Cmd-A) can often be very useful when used outside of only selection 'all cells in the grid'.

A. Select-All within a cell that is not part of a contiguous range should select all cells in the grid. Here is an example: https://gyazo.com/c9b047a221ad8f964854211954fe0186.
B. Select-All within a (non-Table) range, where there is a selection of one or more rectangular cells, should including any cells that are diagonally touching, and select that rectangular region. Doing it again should select all cells in the grid. For example: https://gyazo.com/0444304b5a83b166636f53e886c559be.
C. Select-All within a table has two variations. If one or more data cells are selected, Cmd-A should select all data, pressing it again it should select the entire table, pressing it again should select the entire grid. If one or more cells within the table are selected, and one or more includes a non-data cell, then Cmd-A should select the entire table. If a table-cell and a non-table cell is selected Cmd-A should select the entire grid. For example: https://gyazo.com/a050003df14b1dcff5b408e7f8256c83
D. Select-All where there is one or more selected cells entirely within the pivot table, should select the entire pivot table on Cmd-A. Pressing it again should select the entire grid.

Issue #425, opened 2021-11-25
Performance issues


david542542:
These are considerations when working with a large data set in the grid.

A. Selecting a column of data can be incredibly slow if there are, say, 1M+ rows of data. There are a few 'techniques' for doing this:
1 - Only select the data in the visible frame (i.e., the 50 or so rows).
2 - Select the visible frame immediately, and do the full-selection in the background
3 - Select all the data in the column/row (what currently occurs)

If doing 1 or 2, there would need to be another property such as `entire(Col|Row)SelectionDeferred`. Certain successive events, such as an in-grid sort or copy would require that deferred action, but many events would not (such as moving columns, hiding columns, resizing columns, etc.). This also applies to doing a Select-All of course.

Additionally, it might be good to have a threshold of when to defer the selection of data, such as:

`deferSelectionAfterN = 50,000`

2. I will mention this point here to be explicit, but I don't think the above is applicable to row-selections, as I think about 98% of spreadsheets are under 100 (used) columns and probably 99.99% under 1,000 columns (with the possible exception of a pivot-table with an aggregation on cols).

3. This may already be done, but auto-sizing a column may also take a long time if there are a million rows, if it isn't limited to a sample of the first N rows.

---

To test these items it might be good to create an empty spreadsheet with no data, that is about 10M rows x 1K columns and see how these operations work.

shixued:
我尝试了一个临时解决办法，就是当数据量过大时，先把autoResizeColumns设置为true，并且先加载N条数据，然后将autoResizeColumns设置为false，然后再将剩下的数据追加进去。代码类似：

this.grid =canvasDatagrid({autoResizeColumns:true})

      this.grid.style.height = '100%';
      this.grid.style.width = '100%';


      const totalRows = this.gridStoreData.length;
      const initialRows = 20; // 初始行数
      const batchSize = totalRows;   // 每批大小

      // 1. 初始渲染（快速显示）
      const initialData = this.gridStoreData.slice(0, initialRows);
      console.log(" this.grid", this.grid)
      this.grid.data = initialData; // 首次赋值，表格瞬间出现
      let loadedRows = initialRows;
      this.grid.attributes.autoResizeColumns = false;


          // 2. 使用空闲时间分批追加
      const loadRemainingData=()=> {
        if (loadedRows >= totalRows) {
          // this.grid.draw();
          return;
        }
        
        // 使用requestIdleCallback确保不阻塞交互
[...]

Issue #414, opened 2021-11-25
Improve cell selection


david542542:
There are a few ways to improve the selection of data in one or more cells. They are:

A. Selecting/clicking a single cell should not highlight the inner portion of the cell by default. The reason for this is that 'double-clicking-to-edit' the cell would then produce a flickering effect which is a bit odd. For example: https://gyazo.com/a3d37582bf7e06ac1cc2fdef6c2ff693.

![image](https://user-images.githubusercontent.com/909771/143388717-aba56452-7994-467a-a6df-19c1dedeefcc.png)

B. Previous cell selection(s) should be removed onMouseDown rather than (I think?) onMouseUp, for example: https://gyazo.com/38b94e042f014f9c59ae4f8f814bce72.

C. When selecting a range (if the shift or ctrl key is not being pressed) all previous active/selected cells should be removed. For example: https://gyazo.com/255d7463e38212fdf0dcf5a34a21b208.

D. The first cell selected in a contiguous range (that is, anything without the Ctrl-key being held down) should be the one that is active within the selection. Here is an example: https://gyazo.com/990c020b702f46f75f59b1217e1ae06d. This is important because it is the cell that will receive events such as a right-click or editing text. Currently what is being done is: https://gyazo.com/4cc4e333f6e846405edd7f0579182980. 

![image](https://user-images.githubusercontent.com/909771/143389908-8db8032f-f5d5-4c6a-bf46-1a337b077401.png)

I will add the behavior of ctrl/cmd click to a separate ticket as that's a bit tricky.

david542542:
@hangxingliu can this issue be closed?

hangxingliu:
> @hangxingliu can this issue be closed?

I need to confirm it later. I will let you know after i confirmed

PR #498, opened 2022-02-28
Re-implement(Optimize) select function

hangxingliu:
Here are stages of this PR:

- [x] Utility functions and their unit tests
- [x] Replacing all operations on the `self.selections` in current component with new utility functions
- [x] Add intergration test for it
- [x] Add document to explain its menchanism

The visual tests for the utility functions in this PR: <https://3bdhlh.csb.app/>

Related issues:

- #435 
- #425 
- #414

Hi @ndrsn, this PR doesn't need to be merged before all the stages of it are complete. I create this PR  right now because it can let every developers knows the there some core code is changing and let developers have a peek first.


hangxingliu:
Hi @ndrsn , How do you think this breaking changes:

In original `selections` property is a 2D array. The selection state is represented like this: `self.selections[viewRowIndex] = [0,4,5]`;

But in this PR, `selection` property is a 1D array. Its each item is a selection descriptor. like this: `self.selections[0] = {type: SelectionType.Rows, row0: 10, rows1: 20}`.

There is code to export this property to public (https://github.com/TonyGermaneri/canvas-datagrid/blob/master/lib/intf.js#L1896-L1900). 
And there are many places exported the this propert in the event to the user. like this https://github.com/TonyGermaneri/canvas-datagrid/blob/master/lib/publicMethods.js#L434-L439

Because I think it is necessary to change the implementation of `selection` property if we need to resolve issues I mentioned in this PR. This PR can't be merge if you don't accept this breaking change.

Waiting for your reply to continue changing code. Thanks

hangxingliu:
By the way, I think the property named `columnIndex` in here https://github.com/hangxingliu/canvas-datagrid/blob/master/lib/draw.js#L1157 needs to be changed. But it is a breaking change too because the visible cells can be access by public via `grid.visibleCells`

Because in the other place of this component. `columnIndex` is used for indicating the item index in the schema array. So it can mislead the developer to use it in the wrong place.

ndrsn:
> [](https://github.com/hangxingliu)Because in the other place of this component. columnIndex is used for indicating the item index in the schema array. So it can mislead the developer to use it in the wrong place.

Not sure I completely follow — how would you propose we rename it? And would be renaming the `columnIndex` you're pointing at here, or the other location?

hangxingliu:
> Do you think we would be able to have both versions (two names, grid.selections and grid.<something-else> co-existing at the same time

Yep, I think it is a good idea. (Do you have any advise about naming the new property). I can define a getter for the public access. This getter can return an array like what original selection looks like from the new selection and it will print a warning on the console if anyone access it.

hangxingliu:
> Not sure I completely follow — how would you propose we rename it? And would be renaming the columnIndex you're pointing at here, or the other location?

Because for the compatibility, I don't recommend to change it now. Because in the source code. The name `columnIndex` is used for indicating the column index of the view. (Like the variable name `viewColumnIndex `) But somtimes the name `columnIndex` is used for indicating the item index of the schema. (Like the variable name `boundColumnIndex `). I think we can discuss about this part in the future.

ndrsn:
> Do you have any advise about naming the new property

How about `selectionList` ? Since it's temporary until we've deprecated the old interface (and will keep it a private property?), it's not hugely important to get the naming exactly right, I think.

ndrsn:
> > Not sure I completely follow — how would you propose we rename it? And would be renaming the columnIndex you're pointing at here, or the other location?
> 
> Because for the compatibility, I don't recommend to change it now. Because in the source code. The name `columnIndex` is used for indicating the column index of the view. (Like the variable name `viewColumnIndex `) But somtimes the name `columnIndex` is used for indicating the item index of the schema. (Like the variable name `boundColumnIndex `). I think we can discuss about this part in the future.

I _think_ I understand! Ah yes, let's tackle that later — but good observation.

ndrsn:
I do have a question regarding the `RangeSelector` type: why, if any reason, did you opt for `col0/col1/row0/row1` ? It's a bit more verbose, but would `startColumnIndex/endColumnIndex/startRowIndex/endRowIndex` (or sans -index) work in this context as well? My worry is that a name like `col0` or `col1` would lead the user to think "what about `col2` and `col3`?" — whereas start/end imply (in my mind) more that it's a bounded range. Feelings?

hangxingliu:
> It's a bit more verbose, but would `startColumnIndex/endColumnIndex/startRowIndex/endRowIndex` (or sans -index) work in this context as well?

OK, I will replace them to these style after the features be tested. Because I was worry about the naming of properties are too long. I think we can use the naming you mentioned without `-index`.

- `startColumn`
- `endColumn`
- `startRow`
- `endRow`

ndrsn:
> OK, I will replace them to these style after the features be tested. Because I was worry about the naming of properties are too long. I think we can use the naming you mentioned without -index.

Your suggestion sounds good!

Note: A little context regarding this: my thinking regarding has changed in this respect over the years: I used to go for terse/short names, but I now believe code is read more than it is written thus should be optimized for reading and debugging over writing, so (IMHO) the more explicit and descriptivee the variable names the better. It's not a hard rule or anything, but compared to some of the legacy variable names `c` or `s` (from the days before we had bundlers and optimizers), I believe it's an improvement.

hangxingliu:
> I used to go for terse/short names, but I now believe code is read more than it is written thus should be optimized for reading and debugging over writing

haha, same. I remeber I wrote things like a compiled JavaScript code. But after I touched some large projects. I changed my mind like what you said. And I have also started adding more comment on things that are hard to describe by the name. Thanks.

hangxingliu:
Hi @ndrsn , This PR is ready for review and merge. And I will write a document or blog on the next week about this new selection mechanism.

lib/events/index.js
```diff
@@ -1786,63 +1749,13 @@ export default function (self) {
...
-            selection.pop();
-          }
-        }
-      }
-
-      self.selectionBounds = self.getSelectionBounds();
-      self.selectArea(undefined, ctrl);
-
-      self.draw(true);
+      console.log(e.key, x, y);
```
ndrsn, line 1752: Stray `console.log` ?

lib/intf.js
```diff
@@ -1895,6 +1692,11 @@ export default function (self, ctor) {
     },
   });
   Object.defineProperty(self.intf, 'selections', {
+    get: function () {
+      return self.getObsoleteSelectionMatrix();
```
ndrsn, line 1696: Should we warn the user here, of accessing a soon-to-be-deprecated version of `selections` ?

lib/selections/index.js
```diff
@@ -0,0 +1,904 @@
...
+
+    for (const selectedColumn of selectedColumns) {
+      // Make sure the column is not the row header and that the whole column has in fact been selected.
+      if (selectedColumn >= 0 && self.isColumnSelected(selectedColumn)) {
+        if (isNaN(width)) {
+          const column = schema[self.orders.columns[selectedColumn]];
+          self.fitColumnToValues(column.name);
+        } else {
+          self.sizes.columns[selectedColumn] = width;
+          self.dispatchEvent('resizecolumn', {
```
ndrsn, line 162: @velitasali , this seems to directly impact your work on #502 — depending on the merge order of both of these PRs, one or both of you may encounter a merge conflict or lost work; perhaps you could both could take a look to see what the impact is?

lib/selections/index.js
```diff
@@ -0,0 +1,904 @@
...
+  self.unselectRows = (startRow, endRow, supressEvent) => {
+    const result = removeFromSelections(
+      self.selections,
+      normalizeSelection({
+        type: SelectionType.Rows,
+        startRow: startRow,
+        endRow: endRow,
+      }),
+    );
+    if (result && !supressEvent) self.dispatchSelectionChangedEvent();
```
ndrsn, line 272: Minor nit-pick: there is a typo in `supressEvent`; it should be `suppressEvent` (two p-s)

lib/selections/index.js
```diff
@@ -0,0 +1,904 @@
...
+    return result;
+  };
+
+  /**
+   * @param {object} cell This method needs two properties in this parameter: `rowIndex`
+   * and `viewColumnIndex`
+   * @param {boolean} [supressEvent]
+   */
+  self.selectCell = (cell, supressEvent) => {
```
ndrsn, line 281: Same

lib/selections/index.js
```diff
@@ -0,0 +1,904 @@
...
+
+  /**
+   * This function is migrated from the old API,
+   * @param {boolean} [expandToRow]
+   * @returns {any[]}
+   */
+  self.getSelectedCells = (expandToRow) => {
+    const selectedCells = [];
+    const bounds = self.getSelectionBounds(true);
+    console.log(bounds);
```
ndrsn, line 427: Stray `console.log` ?

ndrsn (changes requested):
@hangxingliu , some minor comments. It's a big PR (!) so I'm not sure I managed to covered all of it properly — if not, apologies. My main question here is: apart from our earlier discussion regarding `selections` and `selectionList`, is there any change in the external interface?

The other point is that yours and @velitasali's work is overlapping a bit, we might need to sort that out first before merging one or the other PR.

lib/selections/index.js
velitasali: @ndrsn, `fitSelectedColumns` is in `selections/index.js` now. I can rebase my commits if this is merged first. It is up to you :)

ndrsn: Ah yeah that sounds like it may be easiest, thanks! Lets wait for @hangxingliu's feedback on my review, and then see if we can merge; it'll probably be tomorrow (Tuesday).

hangxingliu:
Hi @ndrsn, Thanks for your review and suggestions. I will modify the code tonight and ping you again

lib/selections/index.js
hangxingliu: @ndrsn @velitasali , I think it is no worry about the conflict on this function. I believe it can be resolved very easy. I have resolved the conflict between this PR and Veli's last PR. And I will modifiy the code in this PR tonight.

lib/selections/index.js
```diff
@@ -0,0 +1,904 @@
...
+   * Selects a column.
+   * @memberof canvasDatagrid
+   * @name selectColumn
+   * @method
+   * @param {number} columnIndex The column index to select.
+   * @param {boolean} toggleSelectMode When true, behaves as if you were holding control/command when you clicked the column.
+   * @param {boolean} shift When true, behaves as if you were holding shift when you clicked the column.
+   * @param {boolean} supressEvent When true, prevents the selectionchanged event from firing.
+   */
+  self.selectColumn = (columnIndex, ctrl, shift, supressEvent) => {
```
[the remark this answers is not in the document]
hangxingliu: lol, quite lot this typo. it is caused by "copy and paste"

lib/intf.js
```diff
@@ -1894,7 +1691,23 @@ export default function (self, ctor) {
...
     },
   });
+  let warnedForObsoleteSelections;
   Object.defineProperty(self.intf, 'selections', {
+    get: function () {
+      if (!warnedForObsoleteSelections) {
+        console.warn(
+          `DeprecationWarning: The property 'selections' is deprecated due to performance issue. ` +
+            `And it will be removed in the future.` +
+            `Please use new property 'selectionsList' instead. ` +
```
ndrsn, line 1701: Great solution!

But:

`selectionsList -> selectionList`

lib/selections/util.js
```diff
@@ -0,0 +1,1319 @@
...
+
+const SelectionType = {
+  UnselectedCells: 0,
+  Cells: 1,
+  Rows: 2,
+  Columns: 3,
+};
+
+/**
+ * As its name
```
ndrsn, line 23: Missing description?

ndrsn (changes requested):
Very minor comments! Looks great otherwise!

hangxingliu:
@ndrsn  Oh, thanks very much. missed these places.

ndrsn:
All good to go! Thanks a lot, @hangxingliu!

Issue #419, opened 2021-11-25
Fill selection down/over


david542542:
It is often useful to fill data down or over. For example: https://gyazo.com/a8d35951d5533cba9e4aef55b5095f6c. While the logic of this should be implemented on the client-side as to how the fill-down/over should work (and these can be quite complex patterns), the grid should allow the front-end to implement this. This would include:

A. For a contiguous selection of data a little indicator on the lower-right of the selection to be able to grab https://gyazo.com/79e27a56bf98cc8a48450077340e4953.

B. The ability to drag down|up|left|right and show an indicator while the selection is being dragged. For example: https://gyazo.com/c951a76c79bc624fc75764fbbb3ebc0f.

C. The ability to a register a function for how to replicate the data in the filled-to region.

D. A default way to show that it's implemented. For example, maybe we can just copy the value(s) at the edge of the selection (for example, if filling-down, grab the value(s) of the cell in the bottom-most row).

![image](https://user-images.githubusercontent.com/909771/143394246-9f86e07a-a971-46d6-a1e0-5b981e4759b2.png)

david542542:
![image](https://user-images.githubusercontent.com/909771/143658646-452a63eb-9a04-44b6-b6e1-8ce153363d72.png)

david542542:
![image](https://user-images.githubusercontent.com/909771/143658657-641f12f3-e02b-4293-bf44-c5f334349e1e.png)

PR #480, opened 2022-01-30
Ability to hide rows/cols

hangxingliu:
Related issue #420 

- The document about this feature: https://github.com/hangxingliu/canvas-datagrid-patch/tree/master/docs/feature-420
- Latest CodeSanbox demo: <https://myujs.csb.app/> 

And this PR doesn't need to be merged now. I am still adding the UI tests and documents for it. I create a PR first for everyone can have a look now. And you can leave messages here if you find any bugs with this new feature. Thanks

PR #500, opened 2022-03-04
Improve selection and paste

velitasali:
Hi,

Fixes issue #419.

- [x]  Fill data down or over.
- [x]  Add a little indicator on the lower-right of the selection to be able to grab.
- [x] Show an indicator when the handle on the lower-right of the selection is grabbed and moved.
- [x] Ability to a register a function for how to replicate the data in the filled-to region.
- [x] A default way to show that it's implemented.


ndrsn:
Hey @velitasali , thanks for this — looks great so far!

velitasali:
@ndrsn Thanks. I might refactor these after the rest of the features is ready.

velitasali:
Hi @ndrsn,

Could you look at the current state of MR and give feedback if possible? I think the first three should be ready, but I am unsure about what the last two items should do. The fourth commit should address the 4th item, but I might have understood it wrong. 

Thanks.

lib/events/index.js
```diff
@@ -2042,6 +2101,11 @@ export default function (self) {
 
     return rows.length;
   };
+  self.pasteDataReplicate = function (data, dataType, remaining) {
```
ndrsn, line 2104: This might not be entirely finished yet (so ignore this comment if so), and I may be misunderstanding the intent here, but this looks like a function intended to provide the values for the cells that are being 'filled down' (in Excel parlance) — if that's so I don't think `pasteDataReplicate` would be an entirely proper name, because we could, as users, fill down just by dragging the handle, not even pasting the data.

That said, I believe we would need to provide the user an ability to override the default fill down behavior (if we even want to specify a default fill down behavior — what do you think?) using a configurable attribute in the form of a function, or maybe an iterator or (perhaps better) a [generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators#generator_functions). 

See the `Parameters & Attributes` in the [documentation](https://canvas-datagrid.js.org/) for examples of attributes.

To illustrate, I'm thinking something along the lines of:

```js
const grid = canvasDatagrid({
  fillDownFunction: function getFillDownData(data, dataType, ...) {
     return [1, 2, 3];
  }

  // or:

  fillDownFunction: function* makeFillDownIterator(data, dataType, ...) {
    if (isNumericData(data) === false) {
      return []; // we cannot produce the next value if the it's not a numeric sequence
    }

    let lastValue = data[data.length - 1];
    let newValue = lastValue + 1;

    while (true) {
      yield newValue++;
    }
  }
})
```

We could determine whether `fillDownFunction` is a normal function or a generator function by checking its constructor, and acting accordingly.

Note: I'm not entirely sure what arguments would be passed to such a function, hence the `...` notation — but I'd imagine the filling would have to behave differently not just based on the data is being pasted or filled down, but _where_ in the sheet it is being pasted or filled down, so something like `columnIndex/rowIndex` (and `viewRowIndex/viewColumnIndex`) would probably be useful to pass along. Selection size/bounds, as well, maybe?

If the number of positional arguments to such a function becomes too large (let's say more than 3 or 4), I would propose we use an object instead (whatever its contents):

```ts
interface FillDownFunctionProps {
  // ...
}

fillDownFunction: function *makeFillDownIterator(props: FillDownFunctionProps) {
  // ...
}
```

This is all a bit loosely sketched out / thinking out loud — what do you think?

ndrsn (changes requested):
@velitasali, thanks for this! Love the distinct, clearly delineated commits, the code looks great — this was a pleasure to review.

I have a few comments here and there, let me know what you think about them?

lib/events/index.js
```diff
@@ -2109,7 +2131,9 @@ export default function (self) {
       const row = self.selections[rowIndex];
       if (row) {
         if (row[0] === undefined) return;
-        startCell = [rowIndex, row[0] < 0 ? row[1] : row[0]];
+        let isNumbersRowSelected = row[0] < 0;
```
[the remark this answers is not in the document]
velitasali: I now see `isNumbersRowSelected` might not present what it can be. Sorry, it was my time looking at the code there. I will use `row[0] < 0` as an inline statement instead.

lib/events/index.js
```diff
@@ -2190,5 +2213,76 @@ export default function (self) {
...
+    } else {
+      self.selectionHandleOverlay.snap = undefined;
+    }
+  };
+  self.stopSelectionHandleMove = function (e) {
+    if (!self.selectionHandleOverlay.handle) {
+      return false;
+    }
+
+    var overlay = self.selectionHandleOverlay;
```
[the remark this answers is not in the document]
velitasali: I see. Sorry, I will replace those.

lib/defaults.js
```diff
@@ -67,7 +67,6 @@ export default function (self) {
       ['scrollPointerLock', false],
       ['scrollRepeatRate', 75],
       ['selectionFollowsActiveCell', false],
-      ['selectionHandleBehavior', 'none'],
```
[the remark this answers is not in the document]
velitasali: > I can think of use cases where the grid is read-only

Is there a variable/function to check whether the document is in read-only mode. Or more specifically, would `args.editable` be the only factor that alters that behavior? If so, I can use that. 

Btw, I was already working to bring this option back, and I think we can keep the name, which can also preserve the compatibility. The options could be `none`, `single` (bottom right corner only and default), and `all` (all corners when possible). What do you say? If a different name is preferred, I can use that as well.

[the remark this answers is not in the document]
ndrsn: We check `self.attributes.editable` in `beginEditAt` and the `paste` event, so I think we could follow that pattern here as well.

As for name: I think it's good to keep the current name, yes, and then we can provide a deprecation/upgrade path for in future releases. Regarding options: I'm with you on 'single' , but I'm not sure what `all` would do? Does that mean you render a handle in all corners (not opposed, but sounds kinda weird?), or does that allow filling  in all directions (down/up/right/left) ? If the latter, are there examples of cases where you'd want to limit the direction of filling?

lib/events/index.js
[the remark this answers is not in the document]
ndrsn: No apologies necessary :-)

lib/defaults.js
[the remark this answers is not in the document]
velitasali: > We check `self.attributes.editable` in `beginEditAt` and the `paste` event, so I think we could follow that pattern here as well.

Thanks. That sounds easy. I think hiding the selection handles should do the trick here.

> Regarding options: I'm with you on 'single' , but I'm not sure what `all` would do? Does that mean you render a handle in all corners (not opposed, but sounds kinda weird?)

I used `multiple` instead (`all` that sounded weird to me as well :smile:)

> does that allow filling in all directions (down/up/right/left)

I am not sure what you mean by *filling*. It behaves the same, but only the dragging point is different.

> If the latter, are there examples of cases where you'd want to limit the direction of filling?

I might be missing something here. Could you explain what you mean by limiting?


Thanks for the descriptive reply btw :)

velitasali:
Btw, I put selection handle events into the `self.mousemove` to have the ability to autoscroll and not replicate, which needed to me to put a long `if` statement that disallows others functionalities that come before autoscroll. I am aware touchMove already replicates that behavior, so I was wondering your opinion on that. Should I add a new listener for mousemove and replicate, or should I keep it as it is?

lib/defaults.js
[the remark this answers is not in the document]
ndrsn: > I am not sure what you mean by filling. It behaves the same, but only the dragging point is different.

By 'filling' I mean the behaviour when you drag the handle beyond the extent of the selection:

CleanShot 2022-03-10 at 14 39 44

> Again, I might be missing something here. Could you explain what you mean by limiting?

Well, I think you're not so much missing something here, but rather that my confusion is confusing :-) I wasn't sure what `all` or `multiple` actually _meant_ in terms of user experience — does it simply render the handles in all four corners? (And if so, why — does it allow for more capabilities than having the handle in the lower right, as Excel and Google Sheet have it?)

ndrsn:
> Btw, I put selection handle events into the `self.mousemove` to have the ability to autoscroll and not replicate, which needed to me to put a long `if` statement that disallows others functionalities that come before autoscroll. I am aware touchMove already replicates that behavior, so I was wondering your opinion on that. Should I add a new listener for mousemove and replicate, or should I keep it as it is?

Not sure I follow (this isn't code I've ever touched myself) — and this doesn't seem to be showing in the diff just yet, I believe; do you have a line number where I can see the changes? Either way, I probably don't have a strong opinion about it. @hangxingliu is this something you've touched, or have an opinion about?

lib/events/index.js
```diff
@@ -602,163 +602,167 @@ export default function (self) {
...
         };
-        if (self.dragStartObject.columnIndex === -1) {
-          sBounds = self.getSelectionBounds();
-          dragBounds.left = -1;
-          dragBounds.right = s.length - 1;
-          dragBounds.top = Math.min(sBounds.top, cell.rowIndex);
-          dragBounds.bottom = Math.max(sBounds.bottom, cell.rowIndex);
-          if (dragBounds.top < 0) dragBounds.top = 0;
-        }
+        if (self.movingSelectionHandle) {
```
velitasali, line 605: I was talking about this line, which takes over the `mousemove` event when a selection handle is being dragged.

hangxingliu:
> Not sure I follow (this isn't code I've ever touched myself) — and this doesn't seem to be showing in the diff just yet, I believe; do you have a line number where I can see the changes? Either way, I probably don't have a strong opinion about it. @hangxingliu is this something you've touched, or have an opinion about?

I will take a look later and will give my opinion in here

lib/defaults.js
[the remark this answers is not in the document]
ndrsn: In that case, I suggest we keep the behavior currently in place, where any value other than 'none' _enables_ the selection handle.

That will keep it backwards compatible, and then in a future release we can rename this attribute to something more fitting, like a boolean `enableSelectionHandle` _and_ set the default to _true_. How does that sound?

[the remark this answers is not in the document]
velitasali: Yeah, that sounds more reasonable. One thing I should note is the code for selection overlay will keep supporting different handles. Would that be an issue?

hangxingliu:
Hi, @ndrsn  Could you grant this pull request the permission for running the workflow. So we can know can the new code pass the basic tests firstly.

lib/events/index.js
velitasali: I converted it to an event-based format, and moved scroll logic outside `self.selecting` behavior. Everything works fine now.

hangxingliu:
Hi @ndrsn ,I think @velitasali put the logic to individual function is good. This new function may can invoked from other event handler, so it can avoid long if statement and duplicate code.

https://github.com/TonyGermaneri/canvas-datagrid/pull/500/commits/042830fdc3128fb06632eee0a0ae37122e4a9612#diff-3f7a9934546ab5234148ac8c9d607d51b25c0fd156c9f0353d8b3f9d4385b74eR2220-R2270

``` javascript
self.selectionHandleMove = function (e) {
```

lib/events/index.js
velitasali: Hi,

I passed an object instead, and I think it receives all the necessary arguments to easily transform cell data before placing it. What do you think?

velitasali:
This is almost done. I am going to write a few test cases, so I drafted this for now. 

Btw, I realized I had misunderstood some things, so it took me a little to get things right. Now it should work similarly to Google Sheets. Feel free to review.

velitasali:
Hi @ndrsn, one of the tests *I added* is failing, but it is unrelated to the MR and I thought I might get some help for it. Other than that, this is ready for a full review. Let me know what you think.

test/util.js
```diff
@@ -250,6 +250,16 @@ export function touchmove(el, x, y, bbEl) {
     changedTouches: [{ x: 0, y: 0 }],
   });
 }
+export function handlemove(grid, xDiff, yDiff) {
```
velitasali, line 253: This is the cause of the failing test. This should move the handle, but somehow `mousedown` doesn't receive `self.dragMode`. I think the initial `mousemove` should resolve that, but it doesn't. Do you know any other reliable way to fix this?

velitasali: Turns out the handle wasn't visible, so mouse events were failing. Now the test should pass.

lib/events/index.js
```diff
@@ -2036,11 +2175,11 @@ export default function (self) {
       });
     });
 
-    self.dispatchEvent('afterpaste', {
+    self.dispatchEvent('afterfilldown', {
```
ndrsn, line 2178: This feels off to me. Not every paste will be a fill down, will it? Sometimes you're just pasting cells, and sometimes you're just filling cells, and not even pasting the data. But if you emit this event here, you're signalling to the user that every paste you do is _also_ a filldown event, which seems incorrect to me, no?

As an aside, I should note regarding terminology, that "fill down" is only one direction, one can also [fill up, or left, or right](https://support.microsoft.com/en-us/office/fill-a-formula-down-into-adjacent-cells-041edfe2-05bc-40e6-b933-ef48c3f308c6), so to be more correct this event would have to be called `afterfill`:

Fill

(I should have been a bit sharper about correct naming earlier on, apologies for that!)

But getting back to the main point: I feel like two things are getting mixed up here which to me seem related but also feel distinct: pasting and filling data.

To me, _pasting_ is when we just take whatever is in the copy buffer and overwrite the same number cells in the grid with the pasted data:

CleanShot 2022-03-14 at 04 02 31

 _Filling_ is when you need to 'generate' values based on one or more cells in either the currently selected cells or in the copy buffer:

CleanShot 2022-03-14 at 04 06 04

I say 'generate' because these values can either all be the same (as above) or based on some inferred pattern, for example repeating or incrementing:

CleanShot 2022-03-14 at 04 08 27

Notice how the first column is incrementing. The exact logic for generating values is something the user-passed fill function would have to determine how to handle. But in neither of these past two examples am I pasting data, so I would not be expecting an `afterpaste` event.

It gets murkier, though, when you're pasting data into a selection that's larger than the size of your copy buffer:

CleanShot 2022-03-14 at 04 04 20

In this example, we only have two copied cells in our buffer, but we're pasting into an 11x2 selection. I'm not entirely sure how to view that: is that simply a paste action, a fill action, or a paste-then-fill?

_To me_, intuitively, I think it's the latter: we're pasting two cells (the first row of the selection), and then filling _down_ (in this case) the remaining ten rows. So I'd expect both an `afterpaste` and an `afterfill` event.

If so, then I think the code could be structured a little bit simpler, in that the paste function should do what it has been doing, and only after pasting the full copy buffer needs to determine whether any filling is needed as well (i.e. when the selection is larger than the copy buffer bounds). This way you'd be keeping the filling-logic separate from the pasting-logic.

Would love to hear your thoughts on this!

lib/events/index.js
```diff
@@ -1958,56 +1987,166 @@ export default function (self) {
...
-            newRowData[columnName] = existingRowData[columnName];
-            continue;
+          var cellData = isPasting
+            ? cells[colIndex].value.map((item) => item.value).join('')
+            : cells[colIndex];
+          const existingCellData = existingRowData[columnName];
+
+          if (cellCallback) {
+            newRowData[columnName] = cellCallback(
+              Object.assign(fillDownArgs, {
```
ndrsn, line 2123: I'm pretty sure we can use the spread operator here to make our life a bit easier, as well as the object property shorthand (not having to do `{ foo: foo }` but just `{ foo }`:

```js
newRowData[columnName] = cellCallback({
  ...fillDownArgs,
  column,
  columnIndex: startColIndex + colPosition,
  existingCellData,
  // etc
});
```

test/editing.js
```diff
@@ -523,6 +524,137 @@ export default function () {
...
+      data: [
+        {
+          field1: 'Value 1',
+          field2: 'Value 2',
+          field3: 'Value 3',
+        },
+        {},
+      ],
+      fillDownAttributes: {
+        cellCallback: function (args) {
```
ndrsn, line 591: I feel like I'm missing something, but: what are the _other_ properties of `fillDownAttributes` ? I'm only ever seeing `cellCallback` and no other properties, it seems?

ndrsn (changes requested):
@velitasali — I feel like we're almost there! Some technical stuff, but my main interest right now is getting the public interface right: which events to emit when and how to pass the public fill function

Looking forward your input!

lib/events/index.js
```diff
@@ -1958,56 +1987,166 @@ export default function (self) {
...
+      selections,
+    );
+
+    self.dispatchEvent('afterpaste', {
+      cells: affectedCells,
+    });
+
+    return rows.length;
+  };
+  self.fillDown = function (
```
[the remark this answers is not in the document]
velitasali: Yes, it makes sense.

lib/defaults.js
```diff
@@ -67,7 +67,7 @@ export default function (self) {
...
+      ['selectionHandleBehavior', 'single'],
```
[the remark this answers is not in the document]
velitasali: Got it.

lib/events/index.js
velitasali: I totally agree. I should have given more thought to the paste data behavior. However, I think we should agree on the terminology used here. 

I believe we should consider pasting a form of filling data because the logic is almost the same. The confusing part might be the name `fillDown`, which we can rename to `fill` to make it more suitable for pasting. 

Also, separating pasting from filling logic might duplicate the work, since, in the case of pasting, the difference between overfilling or just filling is whether we are in selection bounds or not. That is also the reason why `fillDown` function have `source` as an argument. 

If the source is a paste event, and we have enough data to fill in the selection bounds, we consider that a filling operation, and once the data insufficient, we go on to behave like it is an overfilling one. One thing we can avoid doing here is invoking the cell callback when we are just filling. I think it should become less confusing if I don't do that, unless we are overfilling. 

That being said, you have given a great input here, so I will be some making changes like fixing the `afterfilldown` event.

lib/events/index.js
velitasali: Got it. Thanks!

test/editing.js
velitasali: Yes, it seemed like it is a common in the source code, and I also thought `filling` may get additional attributes. Do you think that is unnecessary? I think I can rename it to something like `fillOverCallback` or something.

ndrsn: True! I think the difference is that in the other cases there's more than one attribute being used, and I wonder if we (will) need more than just the callback function: I feel like all the logic is contained in that one function, which is under the user's control. There _might_ be an argument to be made for parameters to control how/when the fill function should be called, but I cannot think of any, and I don't see the use case for it right now in this PR.

Any ideas?

velitasali: I think, for the sake of compatibility, using an umbrella attribute for the filling behavior might be a good idea. Right now, `cellCallback` is invoked for the filling of each cell, hence the name. But in the future, we might want to add a function that completely overrides the fill behavior, for instance.

Btw, cell-based callback might sound like we have less flexibility, but it is simpler to work with, and I think `args` provides all the necessary input for manipulating the data easily.

velitasali:
eslint freaks out about property spread notation that I added. Is there a reason why that should happen?

hangxingliu:
> eslint freaks out about property spread notation that I added. Is there a reason why that should happen?

Hi @ndrsn , we need to change the the eslint config to support spread notation (changing the value of the property `ecmaVersion` to `latest`):

``` javascript
{
  "root": true,
  "extends": [
    "eslint:recommended",
    "prettier"
  ],
  "parserOptions": {
    "ecmaVersion": "latest",
    "sourceType": "module"
  },
```

test/editing.js
ndrsn: > Btw, cell-based callback might sound like we have less flexibility, but it is simpler to work with, and I think args provides all the necessary input for manipulating the data easily.

I think I agree with this: the fill function should simply receive all the arguments such as cell position, and any other data it needs to determine the current cell's value.

> I think, for the sake of compatibility, using an umbrella attribute for the filling behavior might be a good idea.

This I'm not so sure about — compatibility with what, other arguments? In that case: I feel like we're in a process of gradually improving the public interface (hopefully not breaking anything in the process :), so some inconsistencies are allowed.

And for me, the fact that there are no other attributes than `cellCallback` is an indication that `fillAttributes` is not a great fit. Simpler for documentation and explaining the feature would be a single attribute, e.g. `fillCellFunction` or `fillCellCallback` — to me that feels immediately clear what its intent and shape is.

But I admit, it's a bit of a balancing act. If you feel strongly about `fillAttributes` I'll accept it, but I'm just not entirely convinced yet :-)

hangxingliu:
> eslint freaks out about property spread notation that I added. Is there a reason why that should happen?

Hi @velitasali , I think you can just change the config of the eslint to fix it: changing the ` ecmaVersion` to `12`, like this:

``` javascript
{
  "root": true,
  "extends": [
    "eslint:recommended",
    "prettier"
  ],
  "parserOptions": {
    "ecmaVersion": 12,
    "sourceType": "module"
  },
```

test/editing.js
velitasali: Thanks for the reply :) I don't have a strong preference either, and as you said, so we can go with `fillCellCallback` since there isn't a certainty that `fillAttributes` will ever be expanded upon :)

velitasali:
Thanks, @hangxingliu and @ndrsn.

velitasali:
Hi @ndrsn, I have pushed some new changes (not rebased yet), let me know what you think :)

lib/docs.js
```diff
@@ -106,6 +106,7 @@
  * @param {number} [args.touchZoomMin=0.5] - The minimum zoom scale.
  * @param {number} [args.touchZoomMax=1.75] - The maximum zoom scale.
  * @param {number} [args.maxPixelRatio=1.5] - The maximum pixel ratio for high DPI displays.  High DPI displays can cause sluggish performance, this caps resolution the grid is rendered at.  Standard resolution (e.g.: 1920x1080) have a pixel ratio of 1:1 while higher resolution displays can be higher (e.g.: Retina display 2:1).  Higher numbers are sharper (higher resolution) up to the max of your display (usually 2), lower numbers are lower resolution, down to 1.  It might be fun to set a value lower than 1, I've never done it.
+ * @param {object} [args.fillCellCallback] - The callback function that is called when a cell is filled.
```
ndrsn, line 109: I had to look up how to describe a function `@param` (is `Function` or `function`?), but then discovered you can describe [@callback types](https://jsdoc.app/tags-callback.html) and refer to them by name. For example:

```
@callback fillCellCallback
@param {object} args

// and then later:

@param {fillCellCallback} [args.fillCellCallback] - The callback function that is called when a cell is filled.
```

I don't _exactly_ know how that'll work within the context of `docs.js`, because it feels like all those params belong to the same thing, and I'm not sure what happens when you introduce a `@callback` and then some new `@param`s on subsequent lines — those _should_ belong to the callback, but any other params below shouldn't... no time to explore right now, so just leaving this as a note here. (You don't need to change anything, I'm just thinking out loud — I think a lot of this will be fixed if start introducing TypeScript annotations, like [@hangxingliu is doing in his pull request](https://github.com/TonyGermaneri/canvas-datagrid/pull/498), and using _those_ as basis for documentation)


Also, next step would be to describe the shape of `args` passed to the callback, because `{object}` isn't very descriptive, obviously.

ndrsn (changes requested):
This looks great! A few minor comments, but overall I'm satisfied with the public/external interface ☺️ 

One final request, though: would you be able to put a version of this code online, such as codesandbox? This makes it easier for us to test and click around a little, just to make sure everything's working properly. There's an example of such a codesandbox [here](https://github.com/TonyGermaneri/canvas-datagrid/pull/480).

(Note: I'm working on automating this for pull requests, so in the future you'd only have to include an HTML file in an `examples/` directory, but I haven't gotten that working yet...)

lib/intf.js
```diff
@@ -2057,6 +2058,14 @@ export default function (self, ctor) {
       self.cellGridAttributes = value;
     },
   });
+  Object.defineProperty(self.intf, 'fillCellCallback', {
+    get: function () {
+      return self.fillCellCallback;
+    },
+    set: function setFillDownAttributes(value) {
```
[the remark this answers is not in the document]
velitasali: Thanks, forgot to change that :)

lib/events/index.js
```diff
@@ -2062,6 +2205,27 @@ export default function (self) {
     }
     return -1;
   };
+  self.getSelectionIndex = function () {
```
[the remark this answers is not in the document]
velitasali: I don't think this can achieve the same functionality, since it won't stop unless it iterates through all the elements, even though it should stop with the first failure of `row[0] !== undefined`. Is there a function that does that?

lib/events/index.js
ndrsn: Just wanted to get back to you on this: 

> I believe we should consider pasting a form of filling data because the logic is almost the same. The confusing part might be the name fillDown, which we can rename to fill to make it more suitable for pasting.

Agreed!

> Also, separating pasting from filling logic might duplicate the work, since, in the case of pasting, the difference between overfilling or just filling is whether we are in selection bounds or not. That is also the reason why fillDown function have source as an argument.

My intuition tells me there's a larger difference than just that (for instance, we need an if-statement now to check if we've filled any cells to prevent emitting an `afterfill` event when we're pasting), but I'm having a hard time putting this intuition into words — so as long things are testable and debuggable and in the future relatively easy to refactor, which they now seem to be, it's good with me — I'll resolve this conversation!

lib/events/index.js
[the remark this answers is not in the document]
ndrsn: Ah yes, I _did_ misunderstand! I saw the `break` as `continue`, but of course it's the opposite. In your version, it returns `null` if it encounters an empty-ish (?) row _unless_ it first encounters a non-empty-ish row.

Yeah, I don't have a succinct version for that :-)

lib/events/index.js
velitasali: I just saw this. I want to say I no longer think filling should be considered a form pasting because other platforms don't do that either (guessing from the screenshot you sent). Sorry, I should have written this earlier.

lib/events/index.js
```diff
@@ -2042,6 +2043,148 @@ export default function (self) {
 
     return rows.length;
   };
+  self.insert = function (args) {
```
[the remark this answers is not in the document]
velitasali: The first one seems fine. I applied it.

velitasali:
Here is the demo of this MR: https://codesandbox.io/s/datagrid-demo-v2w6j0?file=/demo.js 

You can observe the changes by pasting data or dragging the handle. Also, the behavior can be configured by tweaking `fillCellCallback` inside **demo.js**.

lib/events/index.js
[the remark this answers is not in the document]
ndrsn: Ah yeah I think you're right: self.selections is a sparse array, I seem to remember, so there might undefined values (hence `if (row) {`

Resolved!

lib/docs.js
ndrsn: Looks great!

ndrsn:
> Here is the demo of this MR: https://codesandbox.io/s/datagrid-demo-v2w6j0?file=/demo.js

Excellent, thank you so much — this makes testing so much easier.

I found a quirk or two. I'm not sure if it's a bug or just my expectation that's off:

CleanShot 2022-03-15 at 13 01 56

When filling down, and then attempting to fill up with the unaltered selection — it fails. Not sure why. It also does not always seem to draw the overlay.

lib/docs.js
```diff
@@ -256,6 +257,32 @@
...
+ * function sets the new data of the cell.
+ * @function
+ * @name canvasDataGrid.fillCellCallback
+ * @param {object} args - Arguments that are passed to the function.
+ * @param {array} args.rows - Data that is being filled into all the cells, including this one.
+ * @param {string} args.direction - Direction of the fill operation, which can be 'horizontal' for x axis, 'vertical' for y axis, or 'both' when it goes both ways, e.g., when pasting unevenly.
+ * @param {object} args.newRowData - New row data that is going to replace 'existingRowData'. The result of this function will be kept in this object.
+ * @param {object} args.existingRowData - Existing data of the row. This will be replaced with 'newRowData'.
+ * @param {number} args.rowIndex - Index of the row on the grid.
+ * @param {number} args.rowPosition - Current position/index of the 'rows'. Starts with '0'.
```
ndrsn, line 271: Is this `rowPosition` a relative position, relative to the start of the selection?

If so, I'd like to suggest we call this `rowOffset` — in my mind `rowPosition` implies a more absolute location, `rowOffset` indicates its relative nature. Same for `columnPosition -> columnOffset`

lib/docs.js
```diff
@@ -256,6 +257,32 @@
...
+ * @param {array} args.cells - All the cells, including this one, in the row. This is the same as accessing 'rows' using 'rowPosition'.
+ * @param {boolean} args.reversed - Whether the fill operation is applied in the reverse order, meaning if horizontal, we are going from right to left, and if vertical, we are going from bottom to top.
+ * @param {boolean} args.fillingRow - Whether the current row is being filled. If the 'direction' is 'vertical', this will always be true, or sometimes true if the 'direction' is 'both'.
```
ndrsn, line 274: This is something I've been mulling for a bit, and it's yet to appear in a contributing style guide anywhere, but I would like to propose that for newly introduced `true/false` variables, where it makes sense and is possible, we name them along the lines of `hasFoo` or `isBlah` or `<verb>Something`, such that `reversed` becomes `isReversed` and `fillingRow` becomes something like `isFillingRow` or `fillsRow` — as a clearer indication that it's a boolean flag. Now for instance, reading `fillingRow` it seems to that's it's the row itself, not a boolean. Turning it into `isFillingRow` or `fillsRow` makes it more obvious, I feel.

What do you think?

ndrsn (changes requested):
I've left a few minor comments regarding naming, but it feels like we can get this merged soon ☺️

lib/docs.js
velitasali: The funny thing is `rowPosition` and `columnPosition` are not provided to the callback, as I already removed them, thinking they were confusing. Apparently, I also forgot to remove them in the docs. So, thanks for catching that :smile:

velitasali:
> I found a quirk or two. I'm not sure if it's a bug

Thanks, I was able to reproduced that and fixed it.

velitasali:
Updated the demo as well.

mdebrauw:
Hey @velitasali, this looks awesome, great work! 

I have been doing some testing on the demo that you've posted and found an issue. 

When I copy a single cell from an Excel or Google sheet to a single cell in CDG the value does not get copied. When copy/pasting a range of cells it works fine. Single cell copy/paste does work in the current version, so I *think* this PR is breaking that (maybe we need to add a test for this?). 

Here's a gif which hopefully clarifies it:

CleanShot 2022-03-15 at 15 27 59

velitasali:
> Hey @velitasali, this looks awesome, great work!

@ndrsn, thank you :)

> When I copy a single cell from an Excel or Google sheet to a single cell in CDG the value does not get copied. When copy/pasting a range of cells it works fine. Single cell copy/paste does work in the current version, so I _think_ this PR is breaking that (maybe we need to add a test for this?).

I was aware of this and forgot to bring it up, so thanks. 

This issue is present in the `master` branch as well, which can be seen here: https://canvas-datagrid.js.org/demo.html. I thought I was using it wrong, so I didn't touch it. Let me know if I should do something about it as well :)

mdebrauw:
Hi @velitasali , hmm, you are right, this issue already exists. It doesn't happen consistently (not sure yet why). Is it a simple fix you think? Otherwise we can raise an issue and fix in separate PR.

ndrsn:
> Hi @velitasali , hmm, you are right, this issue already exists. It doesn't happen consistently (not sure yet why). Is it a simple fix you think? Otherwise we can raise an issue and fix in separate PR.

I'm keen on getting this merged, and if it's a bug that was already present and not directly related to the feature at hand, I say we fix this in a separate commit.

ndrsn:
I think I've discovered another bug, when trying to make the selection smaller the selection seems 'stuck' at the larger size:

CleanShot 2022-03-15 at 16 13 33

Contrast with Excel:

CleanShot 2022-03-15 at 16 15 48

velitasali:
> Hi @velitasali , hmm, you are right, this issue already exists. It doesn't happen consistently (not sure yet why). Is it a simple fix you think? Otherwise we can raise an issue and fix in separate PR.

For reference, the offending line is here: https://github.com/TonyGermaneri/canvas-datagrid/blob/c4d405c88d71d5341e007c7c690a8bac3b68adf0/lib/events/index.js#L1966

What happens is, if there is only one cell to paste, it tries to do a simple paste (a guess). I think it is there for performance reasons, but I might be wrong, so I might need to know what that achieves to fix it.

> I'm keen on getting this merged, and if it's a bug that was already present and not directly related to the feature at hand, I say we fix this in a separate commit.

:+1:

ndrsn:
> Should we invoke the callback when we are shrinking, and what should we call behavior, still _filling_ or something else?

Good questions — I just now tested this out in Google Sheets, and the behavior there seems different still:

CleanShot 2022-03-15 at 16 27 02

In Google Sheets, using the handle only growing the selection is possible, not shrinking. Trying to shrink the selection does nothing, and does not clear the cells like in Excel.

So, considering the different implementations, I think for me it would not crucial that we support the Excel behavior in this PR: this PR implements filling, and the various edge cases are something we might tackle at a later date. That said, if we _do_ want to support the Excel behavior, I feel like we should be able to configure it to support either the Excel or the Sheets behaviour. So an attribute like `allowShrinkSelection` would be needed, as would `clearCellsWhenShrinkingSelection`. I don't think the fill function needs to be called when shrinking the selection, because we only ever want to clear or not clear the cells — I can't think of a situation where you'd need to _fill_ the cells being excluded from the selection. That's also an answer to your question regarding naming the behavior, I think: we can still call it 'filling'.

_Note: naming of these attributes is still open for discussion, I just blurted out what first came to mind_

velitasali:
@ndrsn Could you test if it restores the previous data (I am unable to right now), like you overwrite a cell by moving over it and then shrinking. If it doesn't, it should be easy to implement. 

> So, considering the different implementations, I think for me it would not crucial that we support the Excel behavior in this PR: this PR implements filling, and the various edge cases are something we might tackle at a later date. 

:+1: 

> So an attribute like `allowShrinkSelection` would be needed, as would `clearCellsWhenShrinkingSelection`.

So, these will remain unused for now, right?

ndrsn:
> > So an attribute like `allowShrinkSelection` would be needed, as would `clearCellsWhenShrinkingSelection`.
> 
> So, these will remain unused for now, right?

Yeah those are only once we try to support both Excel and Sheets behavior.

velitasali:
I pushed another commit. I didn't rebase it to make it easier to see the changes. If you prefer the other way around, let me know. 

Also, I am not sure about `allowShrinkSelection` (now `allowShrinkingSelection`) and the other one, as I don't know how Excel behaves in that regard. Web version doesn't support it either (I can't Excel install right now).

ndrsn:
> @ndrsn Could you test if it restores the previous data (I am unable to right now), like you overwrite a cell by moving over it and then shrinking. If it doesn't, it should be easy to implement.

Sorry, missed this earlier — maybe you already figured it out, but just in case, here's what happens:

CleanShot 2022-03-15 at 21 32 28

So it clears the cell values — no need to remember what the original values were — easier, I think?

velitasali:
> Sorry, missed this earlier — maybe you already figured it out, but just in case, here's what happens:

Thanks anyway. It is strange that it is erasing the data instead of undoing the fill, so that should be easier to implement (in another MR probably). Right now, I am making one last optimization as I am not satisfied with how the code finds the corners. It shouldn't take long. 

Btw, I don't think the checks will pass unless I remove the object spread operator.

ndrsn:
> Btw, I don't think the checks will pass unless I remove the object spread operator.

I've updated the .eslintrc file in `master` — if you cherry-pick that commit (or just change 2017 to 2018), it should stop complaining (🤞)

velitasali:
> I've updated the .eslintrc file in `master` — if you cherry-pick that commit (or just change 2017 to 2018), it should stop complaining (crossed_fingers)

Oh. I didn't realize that. Thanks, will do!

ndrsn:
> [](https://github.com/velitasali)Also, I am not sure about allowShrinkSelection (now allowShrinkingSelection) and the other one, as I don't know how Excel behaves in that regard. Web version doesn't support it either (I can't Excel install right now).

Are you not sure about the naming, or the behavior? I did wonder about your preference for continuous tense (shrinking) over present tense? I had to check the docs, seems like we've not been very consistent there, was wondering if you had an argument for one or the other?

velitasali:
> Are you not sure about the naming, or the behavior? I did wonder about your preference for continuous tense (shrinking) over present tense? I had to check the docs, seems like we've not been very consistent there, was wondering if you had an argument for one or the other?

I meant the behavior there. The video you sent cleared things up, so there is no problem now :)

ndrsn:
@velitasali , this feels to me like it's ready to merge — great stuff! How about you?

velitasali:
> @velitasali , this feels to me like it's ready to merge — great stuff! How about you?

@ndrsn I think tit is ready too. Minor issues (if there are any) can be tweaked with smaller commits. I think the features are complete :+1: 

And, thank you for all the reviews/replies, @ndrsn :100:

velitasali:
I am going to make one small change just so that when the user is in the selection bounds, we don't immediately get rid of the direction to preserve the overlay shape until the direction is recalculated. Nothing grand. Afterwards, it can be merged.

ndrsn:
Released in version 0.4.5! 

🎉 

Thanks a lot for all the hard work, @velitasali !

PR #502, opened 2022-03-17
Improve row/column resizing

### LINKED ISSUE #426: Resizing when multiple rows/cols selected
Sheets/Excel has a nice feature where when multiple rows or cols are selected, resizing or auto-sizing fits to all selected cols/rows. For example: https://gyazo.com/faf402901156d934a5540f2f5d20a94a

### LINKED ISSUE #432: How to handle/show dragging
It would be nice to have two settings for how the drag-to-resize columns/rows is done:

- resizeWhileDragging
- resizeAfterDragged

Here would be a difference between the two, including why I think the second option is often a good option: https://gyazo.com/e72fd0ee64a5e015e3a09a9864179816

velitasali:
Hi, 

Fixes #426.
Fixes #432.

- [x] #426
- [x] #432 

Demo: https://codesandbox.io/s/datagrid-demo-v2w6j0?file=/demo.js:4898-4930.

https://user-images.githubusercontent.com/10258162/159027361-a04a5ea2-e1c3-4c87-af74-6e95c5d51e9d.mp4

## Resizing when multiple rows/cols selected #426

The issue was requesting this for both columns and rows, and it was already available for columns, so I only implemented it for rows. 

One thing I am confused about is the dispatched event here: https://github.com/TonyGermaneri/canvas-datagrid/blob/c70d05b8e73314b3245df271cee7f898580ccf94/lib/intf.js#L597 which repeats the same output in the loop. I think moving it out of the loop makes more sense. Or, changing the `draggingItem` to the column that is being altered. 

## Option for how to display col/row resizing #432 

The issue proposes `resizeWhileDragging` and `resizeAfterDragged`. I think putting one of them should do it. `resizeAfterDragged` seems like a good candidate with its default value set to `false`.


velitasali:
@mdebrauw I think fixing it easier.

As a fix, we can provide the width and the column as additional parameters for `fitSelectedColumns`, which should improve the loop situation. And, do the same for `fitSelectedRows`. What do you think?

velitasali:
Hi, 

I decided to fix two similar issues with this MR. They are already similar and dependent on each other.

mdebrauw:
> searched for the `resizerow` event, and values provided by it seem inconsistent. https://github.com/TonyGermaneri/canvas-datagrid/search?q=resizerow. Shouldn't they always be the same?

Good catch. I think you are right, they should be the same.

On the first question: I think it makes sense to have an event for each column resize, it is the most granular level and allows developer flexibility. To support custom logic the user can either use each individual event or debounce till last event and read `sizes.columns` or `sizes.rows` for a complete picture.

velitasali:
> Good catch. I think you are right, they should be the same.

I am going to open an issue :+1: 

> On the first question: I think it makes sense to have an event for each column resize, it is the most granular level and allows developer flexibility. To support custom logic the user can either use each individual event or debounce till last event and read `sizes.columns` or `sizes.rows` for a complete picture.

Thanks for the reply. I am going to add `columnIndex` as a parameter to all `resizecolumn` events and do the similar for `resizerow`.

ndrsn:
Hi @velitasali , looks great! This looks like it's ready to merge, but:

> Thanks for the reply. I am going to add columnIndex as a parameter to all resizecolumn events and do the similar for resizerow.

Not seeing this reflected in the diffs — is this something you still want to do in this PR?

velitasali:
> Not seeing this reflected in the diffs — is this something you still want to do in this PR?

One of the blockers is https://github.com/TonyGermaneri/canvas-datagrid/issues/503. Also, other `resizecolumn` events might need a change, or, if adding nullable parameters is accepted, I can add row index and column index to both `fitSelectedRows` and `fitSelectedColumns`.

ndrsn:
Ah, I see! If I understand correctly, for these events, can we adapt a gradual approach and add the preferred parameter names where appropriate, while leaving whatever old one is present? For example:

```js
  self.dispatchEvent('resizerow', {
    x: self.resizingStartingWidth,
    y: height,
    draggingItem: self.currentCell,
  });
```
becomes
```js
self.dispatchEvent('resizerow', {
  x: self.resizingStartingWidth,

  // TODO: deprecated, remove in future release:
  y: height,

  rowHeight: height, // <-- new

  draggingItem: self.currentCell,
});
```

Then in a future release we can remove `y` after giving sufficient warning. Or would that complicate things?

velitasali:
> Then in a future release we can remove y after giving sufficient warning. Or would that complicate things?

I think we'd also need to rename `x` to `rowWidth` if I am not mistaken. That said, shouldn't we also add the row/column index to the event data to dispatch in the loop? I think the discussion was around that.

ndrsn:
> I think we'd also need to rename `x` to `rowWidth` if I am not mistaken. That said, shouldn't we also add the row/column index to the event data to dispatch in the loop? I think the discussion was around that.  

I _think_ we're saying the same thing — I'm suggesting not _renaming_ but _adding_ `rowWidth` (and `-Height`) to the current event data

velitasali:
> I _think_ we're saying the same thing — I'm suggesting not _renaming_ but _adding_ `rowWidth` (and `-Height`) to the current event data, and then removing `x/y/etc` at a later date

I think the latest changes should do. I went with `width` and `height` instead of `rowHeight` etc., since `heigth` is coming from `sizes.rows` and `width` from `sizes.columns` (if I understood correctly).

lib/events/index.js
```diff
@@ -958,14 +959,32 @@ export default function (self) {
       y = self.style.minRowHeight;
     }
     if (
-      self.dispatchEvent('resizecolumn', {
+      self.dispatchEvent(resizingColumn ? 'resizecolumn' : 'resizerow', {
         x: x,
         y: y,
+        width: x,
+        height: y,
+        index: resizingColumn
```
mdebrauw, line 967: @velitasali, apologies, late to the party here. The key-name `index` here is not particular descriptive. I think using `columnIndex` and `rowIndex` for `resizecolumn` and `resizerow` events respectively makes it more explicit what you're dealing with as a consumer of the api/events, e.g. `{ ... index: 3 }` vs `{ ... rowIndex: 3 }`. 

What do you think?

ndrsn: I'm inclined to agree! I wonder if we should then always have both `rowIndex` _and_ `columnIndex` present (where for instance `columnIndex: null` if the event is `resizerow`) or only one of the two keys, depending on the type of event.

My instinct says _both_, but I can't quite articulate why.

velitasali: @mdebrauw I am not sure. If it is going to help users to have an easier time, we can go that way (explicitly specifying `columnIndex` or `rowIndex`). Other than that, the current approach should require less boilerplate, imo.

ndrsn: If boilerplate when _writing_ internal canvas-datagrid code is a concern, we could consider a helper function? I very much liked @hangxingliu's approach with [getContextOfSelectionEvent](https://github.com/TonyGermaneri/canvas-datagrid/pull/498/files#diff-4430e2d62644251165d972c62b389ef52b749cbd19c3ee418888833190317801R382), as a way of getting/providing the data needed to emit an event.

```js
self.dispatchEvent('selectionchanged', self.getContextOfSelectionEvent());
```

This should reduce the boilerplate needed, I think. Even so, I tend towards preferring the code more explicit, clearer to debug and easier read for future contributors — but perhaps at the risk of being overly verbose/boilerplate-y. But it's a balancing act, for sure.

velitasali: I see, but I am not sure if there is a need for that here. I mean, is there a function that we can offload the parameters provided by the event like `getContextOfSelectionEvent` does?

ndrsn: Ah right, I see what you mean. Sorry, never mind then :-)

But I don't think I'm understanding the "less boilerplate" argument — we want to be passing the row or column index for each `resizerow` or `resizecolumn` event, correct? So then it makes no difference, in terms of boilerplate, whether it's called `index` instead of `rowIndex` or `columnIndex` ?

velitasali: Now that I think about it, I am also for going with `rowIndex` and `columnIndex` since users can assign the same function to  both `resizerow` and `resizecolumn`. For the index that is not available, we can give -1 or undefined, which (I think) is more predictable.

ndrsn: Ah yes, I didn't even consider using the same function — good one. `undefined` sounds great!

ndrsn:
@velitasali, as far as I'm concerned this looks great to merge! What do you think?

velitasali:
@ndrsn Thanks. I think it is ready too.

ndrsn:
Alright, merging! Holding off releasing a new version until [other PR has been merged, as well](https://github.com/TonyGermaneri/canvas-datagrid/pull/505#issuecomment-1075854489).

PR #515, opened 2022-06-28
Fix `allowColumnReordering: true` not working while `allowRowReordering` is `false`

HitomiTenshi:
Fixes issue #514.

PR #550, opened 2023-03-25
fix sorting on header click

### LINKED ISSUE #261: allowSorting does not disable sorting on header clicks
It does remove the option from the context menu.  This is simple to fix.

check the allowSorting bool just before this line

https://github.com/TonyGermaneri/canvas-datagrid/blob/aee9a2a12d43dda54d615e619b90031f37f32527/lib/publicMethods.js#L642

romanstetsyk:
Fixes #261.

- check the value of the `allowSorting` attribute before sorting.

### DIFF
--- a/lib/publicMethods.js
+++ b/lib/publicMethods.js
@@ -701,6 +701,10 @@ export default function (self) {
     sortFunction,
     dontSetStorageData,
   ) {
+    if (!self.attributes.allowSorting) {
+      self.refresh();
+      return;
+    }
     var f,
       c = self.getSchema().filter(function (col) {
         return col.name === columnName;


ndrsn:
Hey @romanstetsyk, thanks for this! This change [causes a test to fail](https://github.com/TonyGermaneri/canvas-datagrid/actions/runs/4521161715/jobs/7963525846?pr=550#step:6:2624), but I'm not sure that test is correct to start with.

So I'm going to merge this soon, but I should tell you that we can't do a new release for this right away because there's another failing test caused by an earlier change which still is under investigation. I'll update this thread once I do a new release.

Thanks again :)

ndrsn:
> This change [causes a test to fail](https://github.com/TonyGermaneri/canvas-datagrid/actions/runs/4521161715/jobs/7963525846?pr=550#step:6:2624)

I believe the actual failure for the test was introduced in #515, so reverted that.

Issue #552, opened 2023-04-06
Feature request. Allow pasting tables bigger than grid


romanstetsyk:
### Expected behavior and actual behavior.

Actual: if a table bigger than the grid is pasted, rows and columns that don't fit are cut out.
Expected: new columns and rows are added to the grid to contain all pasted data.

### Steps to reproduce the problem.

Paste data that has more rows or columns than the grid.

### Specifications like the version of the project, operating system, or hardware.
`canvas-datagrid v0.4.6`

PR #553, opened 2023-04-06
Allow adding new rows or columns on paste

romanstetsyk:
Fixes issue #552.

- Add a new attribute, `allowGridSizeChangeOnPaste` (which is `false` by default). If it's set to `true` and pasted data is bigger than the grid, new rows or columns will be created as needed. Please see the screen recording.

https://user-images.githubusercontent.com/25715951/230503165-b5f32a6e-8dbe-44c4-96be-8c94ae87186e.mov

This is slow for big datasets because a) `.addRow()` calls `.refresh()` b) on each cell pasted `.selectCell()` is called. So there's plenty of room for optimization.

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -18,6 +18,7 @@ export default function (self) {
       ['allowSorting', true],
       ['allowGroupingRows', true],
       ['allowGroupingColumns', true],
+      ['allowGridSizeChangeOnPaste', false],
       ['animationDurationShowContextMenu', 50],
       ['animationDurationHideContextMenu', 50],
       ['autoGenerateSchema', false],
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -18,7 +18,7 @@ export default function (self) {
       ['allowSorting', true],
       ['allowGroupingRows', true],
       ['allowGroupingColumns', true],
-      ['allowGridSizeChangeOnPaste', false],
+      ['allowGridExpandOnPaste', false],
       ['animationDurationShowContextMenu', 50],
       ['animationDurationHideContextMenu', 50],
       ['autoGenerateSchema', false],
--- a/lib/docs.js
+++ b/lib/docs.js
@@ -39,7 +39,7 @@
  * @param {string} [args.blanksText=(Blanks)] - The text that appears on the context menu for filtering blank values (i.e. `undefined`, `null`, `''`).
  * @param {string} [args.ellipsisText=...] - The text used as ellipsis text on truncated values.
  * @param {boolean} [args.allowSorting=true] - Allow user to sort rows by clicking on column headers.
- * @param {boolean} [args.allowGridSizeChangeOnPaste=false] - Allow adding new rows and columns if pasted data dimentions are bigger than existing grid dimentions.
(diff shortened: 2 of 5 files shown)


lib/defaults.js
```diff
@@ -18,6 +18,7 @@ export default function (self) {
       ['allowSorting', true],
       ['allowGroupingRows', true],
       ['allowGroupingColumns', true],
+      ['allowGridSizeChangeOnPaste', false],
```
ndrsn, line 21: Proposal, but maybe a nitpick: `allowGridSizeChangeOnPaste` allows for the possibility that it might either expand _or_ contract, so maybe call it `allowGridExpandOnPaste` ?

lib/docs.js
```diff
@@ -39,6 +39,7 @@
  * @param {string} [args.blanksText=(Blanks)] - The text that appears on the context menu for filtering blank values (i.e. `undefined`, `null`, `''`).
  * @param {string} [args.ellipsisText=...] - The text used as ellipsis text on truncated values.
  * @param {boolean} [args.allowSorting=true] - Allow user to sort rows by clicking on column headers.
+ * @param {boolean} [args.allowGridSizeChangeOnPaste=false] - Allow adding new rows and columns if pasted data dimentions are bigger than existing grid dimentions.
```
ndrsn, line 42: `dimentions -> dimensions`

lib/events/index.js
```diff
@@ -2007,7 +2007,7 @@ export default function (self) {
     alwaysFilling = false,
     direction = 'both',
   }) {
-    var schema = self.getSchema();
+    // var schema = self.getSchema();
```
ndrsn, line 2010: Can we remove this line, if we no longer use `schema` here?

lib/events/index.js
```diff
@@ -2073,15 +2083,25 @@ export default function (self) {
...
-          console.warn('Paste data exceeded grid bounds. Skipping.');
-          continue;
+        if (!self.getSchema()[columnIndex]) {
+          if (self.attributes.allowGridSizeChangeOnPaste) {
+            const lastColSchema = self.getSchema()[
+              self.orders.columns[self.getSchema().length - 1]
+            ];
+            const newColSchema = {
+              ...lastColSchema,
+              name: `col${self.schema.length + 1}:${Date.now()}`,
```
ndrsn, line 2093: Should we not be using `self.getSchema()` here? Or rather:

```js
const schema = self.getSchema();

if (!schema[columnIndex]) {
  if (self.attributes.allowGridSizeChangeOnPaste) {
    const lastColSchema = schema[self.orders.columns[schema.length - 1]];
  }
  const newColSchema = {
    ...lastColSchema,
    name: `col${schema.length + 1}${Date.now()}`,
    title: ' ',
  };
  self.addColumn(newColSchema)
}
```

Also, two (three?) other points:

1) can we safely assume that the last column's schema is similar to the newly pasted column? What is you're pasting text and the previous existing column schema defines it as `type: number` ?
2) What is the purpose of the name including the current date? Do we even need to name the new column?
3) Why the empty title?

ndrsn (changes requested):
Great work @romanstetsyk, thanks so much! I have a few minor points, but I really appreciate the tests you've written in addition to the new functionality — really makes things easier to merge :)

PR #586, opened 2025-09-12
feat: Add clearRect before draw background.

Fozei:
When the background color is set with a transparency parameter (such as "transparent" or when the fourth parameter of rgba is not 1), the canvas retains previously drawn content.

Reproduction steps:
1. set `gridBackgroundColor` to `"transparent"`
2. Set the grid.data array to a new array that its length shorter than it was previously.


TonyGermaneri:
Thanks for the fix!

PR #589, opened 2026-09-10
Add opt-in last-column fill behavior

### LINKED ISSUE #474: Contract / Bounty - 3: Make last column width = right of grid, disable last column right side resize and remove 'dead space'
The Reason - I'm looking to use canvas-datagrid more like a regular multi column list than an excel like experience.

The Problem - A user can currently resize the very right hand side of the last column.  This creates a 'dead space' area to the right of the grid. A user can also resize other columns and the last column width (or right side) doesn't fill the empty space of the column width also leaving the 'dead space'.

The fix - Add a new Parameter to mark the last column as autoMaxWidth or something. This would disable a users ability to resize the very right hand side of the last column and also make the columns width always = the right of the grid , thus preventing the visible 'dead space'.

canvas-dg-issue-3

Cauho:
Fixes #474.

Adds a `fillLastColumn` option that expands the last visible column to the usable right edge, prevents resizing that outer edge, recomputes when the grid or option changes, and leaves genuinely overflowing grids scrollable.

Validation:
- `npm test` — 239 passed, 14 skipped; the command also completed both production and debug builds.
- Prettier check passed for the changed library files.
- ESLint reported 0 errors; 12 existing indentation warnings remain in untouched lines.
- The seven remote file blobs match the locally tested files exactly.

Implementation and tests were prepared with Codex assistance.


Cauho:
Hi Tony, just following up on the fillLastColumn implementation for #474. Does this match the behavior you wanted, or is there anything you'd like changed before review? Could you also confirm whether this task is still eligible for the advertised contract/bounty, and the amount and payment method if accepted?

